PR #2, opened 2021-02-16
update plugin with yaml tags

kkeller-cb:
We noticed that the `Account` array was always empty. This pr adds yaml tags to the config structs so it populates.

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -55,14 +55,14 @@ type Provider struct {
 
 type Account struct {
 	ID      string
-	RoleARN string
+	RoleARN string `yaml:"role_arn"`
 }
 
 type Config struct {
-	Regions    []string
-	Accounts   []Account
-	LogLevel   *string
-	MaxRetries *int
+	Regions    []string  `yaml:"regions"`
+	Accounts   []Account `yaml:"accounts"`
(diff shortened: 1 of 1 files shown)


yevgenypats:
@kkeller-cb Thanks! I double tested it now and looks like some of the fields do get populated as long as you specify it in the `config.yml`. I did see that the `RoleARN` does not get populated without the tag indeed.

I think the yaml package uses a snake_case by default for the tags. Anyway to avoid bugs like with `RoleARN` I think it is a good practice to have the tags explicit and merge this (there is one typo in the PR though).

provider.go
```diff
@@ -55,14 +55,14 @@ type Provider struct {
...
 }
 
 type Config struct {
-	Regions    []string
-	Accounts   []Account
-	LogLevel   *string
-	MaxRetries *int
+	Regions    []string  `yaml:"regions"`
+	Accounts   []Account `yaml:"accounts"`
+	LogLevel   *string   `yaml:"log_leve"`
```
yevgenypats, line 64: should be `log_level` not `log_leve`

kkeller-cb: fixed

PR #9, opened 2021-03-04
feat: eks support

James-Quigley:
Adds support for ingesting EKS cluster information. I believe all fields should be captured.

There is no way to run a describe on multiple clusters at once, so I went with the approach to list clusters and describe one by one.

### DIFF
--- a/resources/eks/clusters.go
+++ b/resources/eks/clusters.go
@@ -175,32 +175,38 @@ func (c *Client) transformEKSClusters(values *[]types.Cluster) []*Cluster {
 	var tValues []*Cluster
 	for _, value := range *values {
 		tValue := Cluster{
-			AccountID:                c.accountID,
-			Region:                   c.region,
-			Name:                     value.Name,
-			Arn:                      value.Arn,
-			CertificateAuthorityData: value.CertificateAuthority.Data,
-			CreatedAt:                value.CreatedAt,
-			Endpoint:                 value.Endpoint,
-			PlatformVersion:          value.PlatformVersion,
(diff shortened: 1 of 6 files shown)


resources/eks/clusters.go
```diff
@@ -0,0 +1,252 @@
...
+			EndpointPublicAccess:     value.ResourcesVpcConfig.EndpointPublicAccess,
+			RoleArn:                  value.RoleArn,
+			Status:                   string(value.Status),
+			Version:                  value.Version,
+			Tags:                     c.transformClusterTags(&value.Tags),
+			LoggingConfigurations:    c.transformClusterLoggingConfigurations(&value.Logging.ClusterLogging),
+			PublicAccessCidrs:        c.transformClusterPublicAccessCidrs(&value.ResourcesVpcConfig.PublicAccessCidrs),
+			SecurityGroups:           c.transformClusterSecurityGroups(&value.ResourcesVpcConfig.SecurityGroupIds),
+			Subnets:                  c.transformClusterSubnets(&value.ResourcesVpcConfig.SubnetIds),
+		}
```
yevgenypats, line 198: I think `nil` checks are missing for `value.ResourcesVpcConfig` and `value.Logging` just like you did below.

yevgenypats (changes requested):
Thanks! One comment and good to go!

resources/eks/clusters.go
yevgenypats: Thanks. @James-Quigley Thanks! I think you forgot to delete it from the above struct

yevgenypats:
Merged. Thanks! 🥳

PR #13, opened 2021-03-10
Add concurrent account fetching

yevgenypats:
https://github.com/cloudquery/cloudquery/issues/58

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -34,7 +34,6 @@ import (
 	"github.com/hashicorp/go-hclog"
 	"golang.org/x/sync/errgroup"
 	"strings"
-	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


provider/provider.go
```diff
@@ -194,24 +188,114 @@ var allRegions = []string{
 	"sa-east-1",
 }
 
-func (p *Provider) Fetch(data []byte) error {
+func (p *Provider) validateFetchConfig() error {
+	if len(p.config.Resources) == 0 {
+		p.Logger.Info("no resources specified. See available resources: see: https://docs.cloudquery.io/aws/tables-reference")
+		return nil
```
James-Quigley, line 194: should this be an error instead of nil?

James-Quigley:
Runs just fine and was a significant speed improvement in my testing.

PR #14, opened 2021-03-11
fix: create new retryer configuration for each service

James-Quigley:
fixes issue described in sdk docs where a global retry token bucket was being shared across all services https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/retries-timeouts/#customizing-behavior

This meant that if a single service hit a backoff, then every services subsequent requests would be heavily delayed, making ingestion slow.

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -46,7 +46,6 @@ type Provider struct {
 	db              *database.Database
 	config          Config
 	Logger          hclog.Logger
-	retryOpt        config.LoadOptionsFunc
 	regions         []string
 }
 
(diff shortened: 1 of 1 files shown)


provider/provider.go
```diff
@@ -236,7 +243,9 @@ func (p *Provider) fetchAccount(accountID string, awsCfg aws.Config, svc *sts.Cl
 
 		innerLog := p.Logger.With("account_id", accountID, "region", region)
 		for serviceName, newFunc := range regionalServices {
-			resourceClients[serviceName] = newFunc(awsCfg,
+			cfg := awsCfg.Copy()
```
yevgenypats, line 246: I think that there is no need for this line as it is already copied when `awsCfg` pass to this function (as it's not a pointer).

yevgenypats (changes requested):
Looks good! one minor comment.

PR #19, opened 2021-03-18
fix: bulk insert ec2 for each describe call instead of one by one

### LINKED ISSUE #18: EC2 inserts one by one into database
Majority of EC2 reservations have only a single instance, so https://github.com/cloudquery/cq-provider-aws/blob/main/resources/ec2/instances.go#L590 transforms a single instance and inserts a single instance. For accounts that have a lot of ec2 instances, this really slows down fetching and makes quite a lot of small DB inserts.

Ideally CQ would transform all instances for all reservations, and do a single bulk insert?

James-Quigley:
resolves #18

### DIFF
--- a/resources/ec2/instances.go
+++ b/resources/ec2/instances.go
@@ -587,10 +587,12 @@ func (c *Client) instances(gConfig interface{}) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


yevgenypats:
Good catch. Thanks!

PR #21, opened 2021-03-24
Fetch IAM User data from ListUsers and CredentialReport

### LINKED ISSUE #20: Bug: Deleting an IAM user after credentialreport is generated causes a 404 error
Cloudquery fetches the list of users from the credential report, then calls a get user on every user in the report. If you've deleted a user since the credential report was generated (up to 4 hours previous, as described in https://github.com/cloudquery/cloudquery/issues/44), this causes cloudquery to fail with a 404 error.

Should probably get users from another source, but also not fail if a request 404s

James-Quigley:
Fixes #20.

Gets a list of users from ListUsers as well as a credential report and merges the data, ignoring any unused data from the credential report

### DIFF
--- a/resources/iam/users.go
+++ b/resources/iam/users.go
@@ -3,14 +3,16 @@ package iam
 import (
 	"context"
 	"errors"
-	"github.com/aws/aws-sdk-go-v2/aws"
-	"github.com/aws/aws-sdk-go-v2/service/iam/types"
-	"github.com/aws/smithy-go"
 	"strings"
 	"time"
 
-	"github.com/aws/aws-sdk-go-v2/service/iam"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/iam/types"
+	"github.com/aws/smithy-go"
 	"github.com/gocarina/gocsv"
+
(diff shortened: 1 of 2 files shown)


resources/iam/users.go
```diff
@@ -237,46 +245,57 @@ func (c *Client) transformReportUser(ctx context.Context, reportUser *ReportUser
...
+		}
 	}
 
 	return &res, nil
 }
 
-func (c *Client) transformReportUsers(ctx context.Context, values []*ReportUser) ([]*User, error) {
+func getMatchingReportUser(reportUsers []*ReportUser, user types.User) *ReportUser {
+	for _, reportUser := range reportUsers {
+		if user.Arn == &reportUser.ARN {
```
yevgenypats, line 288: Strange that it worked because looks like you are comparing pointers. It should be the other way around. `*user.Arn == reportUser.ARN`

yevgenypats (changes requested):
Thanks! One comment.

James-Quigley:
I think this should be good to go @yevgenypats

yevgenypats:
@James-Quigley merging! for some reason didn't get the notification of the additional commit.

PR #54, opened 2021-04-30
Allow fetching EBS volumes

bombergio:
With this PR adding new config argument `ec2.ebs_volumes` to fetch EC2 EBS volumes.


resources/ec2_ebs_volumes.go
```diff
@@ -0,0 +1,121 @@
...
+				Name: "encrypted",
+				Type: schema.TypeBool,
+			},
+			{
+				Name: "fast_restored",
+				Type: schema.TypeBool,
+			},
+			{
+				Name: "iops",
+				Type: schema.TypeInt,
```
James-Quigley, line 52:
```suggestion
				Type: schema.TypeBigInt,
```

resources/ec2_ebs_volumes.go
```diff
@@ -0,0 +1,121 @@
...
+				Name: "multi_attach_enabled",
+				Type: schema.TypeBool,
+			},
+			{
+				Name: "outpost_arn",
+				Type: schema.TypeString,
+			},
+			{
+				Name: "size",
+				Type: schema.TypeInt,
```
James-Quigley, line 68:
```suggestion
				Type: schema.TypeBigInt,
```

James-Quigley (changes requested):
A few comments:
- I think types that are int64 should correspond to BigInt?
- I think this is the right schema from the SDK right? https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v0.24.0/service/ec2#Volume If so, I don't see `throughput` in the schema, and it seems like `Attachments` were omitted
- Need to run the generator for the mocks to update the ec2 service mock, and then need to implement the mock test for ec2 Volumes. Follow the instructions in https://github.com/cloudquery/cq-provider-aws/blob/main/adding_a_new_resource.md

bombergio:
You are referencing outdated EC2 SDK - v 0.24.0, while I was using v 1.2.0 (and the same version is used in the [module](https://github.com/cloudquery/cq-provider-aws/blob/main/go.mod#L15) )
As you can see in the [EC2 v1.2.0](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2@v1.2.0/types#Volume), all integer fields are of type `int32` and there is a field name `throughput`.
Thanks for pointing missing `Attachments` and missing mocks, I'll add them now.

resources/ec2_ebs_volumes.go
bombergio: Type is `int32`, so no need to change it to the `TypeBigInt`

resources/ec2_ebs_volumes.go
bombergio: Type is `int32`, so no need to change it to the `TypeBigInt`

yevgenypats:
@bombergio Thanks! Merging.

PR #60, opened 2021-05-05
Route53 hosted zones implementation

fdistorted:

PR #63, opened 2021-05-13
Elbv1

fdistorted:
elbv1 implementation


fdistorted:
I have used  custom type and loaded attributes on fetch phase. Then each attributes field was resolved separately. Attributes could be set in post resolver. I is simpler to read because all fields names are set automatically and structure is preserved.

resources/elbv1_load_balancers.go
```diff
@@ -0,0 +1,547 @@
...
+			getTagById := func(id string, tagsResponse []types.TagDescription) []types.Tag {
+				for _, t := range tagsResponse {
+					if id == *t.LoadBalancerName {
+						return t.Tags
+					}
+				}
+				return nil
+			}
+
+			if (i+1)%20 == 0 || i+1 == cap(loadBalancerWrappers) {
```
roneli, line 329: See my comment in route53 PR #60

resources/elbv1_load_balancers.go
```diff
@@ -0,0 +1,547 @@
...
+	if r.Attributes == nil && r.Attributes.CrossZoneLoadBalancing == nil {
+		return nil
+	}
+	resource.Set(c.Name, r.Attributes.CrossZoneLoadBalancing.Enabled)
+	return nil
+}
+func resolveElbv1loadBalancerAttributesConnectionDrainingEnabled(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+	r, ok := resource.Item.(ELBv1LoadBalancerWrapper)
+	if !ok {
+		return errors.New("not load balancer")
```
roneli, line 431: I saw you defined an error in #60 would we just use it for all errors ?

roneli (changes requested):

resources/elbv1_load_balancers.go
```diff
@@ -0,0 +1,550 @@
...
+	for {
+		response, err := svc.DescribeLoadBalancers(ctx, &config, func(options *elbv1.Options) {
+			options.Region = c.Region
+		})
+		if err != nil {
+			return err
+		}
+
+		for i := 0; i < len(response.LoadBalancerDescriptions); i += 10 {
+			end := i + 20
```
roneli, line 345: +10 or+ 20 which is correct? better to use a const type for this number.

resources/elbv1_load_balancers.go
```diff
@@ -0,0 +1,550 @@
...
+}
+
+// ====================================================================================================================
+//                                               Table Resolver Functions
+// ====================================================================================================================
+func fetchElbv1LoadBalancers(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan interface{}) error {
+
+	c := meta.(*client.Client)
+	svc := c.Services().ELBv1
+	processLoadBalancers := func(loadBalancers []types.LoadBalancerDescription) error {
```
roneli, line 303: why not move this into a function under elbv1_types and pass the svc + load balancers?

roneli (changes requested):
Looks great, 1 minor suggestion left and issue with chunk size increase

roneli (approved):

PR #70, opened 2021-05-19
Fixed Quota error by multiplying services per account and region.

zagronitay:
https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html

https://github.com/cloudquery/cloudquery/issues/94


client/client.go
```diff
@@ -95,15 +97,36 @@ type Services struct {
...
 }
 
+type ServicesAccountRegionMap map[string]map[string]*Services
+
+// ServicesManager will hold the entire map of (account X region) services
+type ServicesManager struct {
+	services ServicesAccountRegionMap
+}
+
+func (s *ServicesManager) ServiceByAccountAndRegion(accountId string, region string) *Services {
```
roneli, line 107: Services*ByAccountAndRegion

client/client.go
```diff
@@ -67,6 +67,8 @@ var allRegions = []string{
 	"sa-east-1",
 }
 
+var defaultRegion = "us-east-1"
```
roneli, line 70: const

roneli (approved):
Looks good, minor suggestions (grammar + const)

yevgenypats (changes requested):
I've tested it and Im not sure it solved the problem. The AWS client should be goroutine safe per documentation also the retrier is automatically copied like all other options on each call. For example see this : https://github.com/aws/aws-sdk-go-v2/blob/main/service/autoscaling/api_client.go#L121 

I suspect this is something else like incorrect retrier configuration in the first place. I've a few ideas but let's discuss this live tomorrow

@roneli @zagronitay

yevgenypats (approved):

PR #85, opened 2021-06-01
AWS WAF V1 implementation

michelvocks:
Implements https://github.com/cloudquery/cloudquery/issues/117


adding_a_new_resource.md
```diff
@@ -12,6 +12,7 @@ As a prerequisite, in [aws-sdk-go-v2](https://pkg.go.dev/github.com/aws/aws-sdk-
 If the service to which the resource belongs has not been used before in cq-provider-aws, there are a few steps that need to be done to configure it.
 
 1. Create the service interface in [client/services.go](./client/services.go)
+   * Don't forget to add new service interface name to the go:generate comment.
```
michelvocks, line 15:
```suggestion
   * Don't forget to add the new service interface name to the go:generate comment.
```

resources/waf_rules.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/waf"
+	"github.com/cloudquery/cq-provider-aws/client"
+	"github.com/cloudquery/cq-provider-sdk/provider/schema"
+)
+
+func WafRules() *schema.Table {
```
roneli, line 12: You are missing [tags ](https://docs.aws.amazon.com/waf/latest/APIReference/API_waf_ListTagsForResource.html)for WafRules, WebAcls and WafRuleGroups.

resources/waf_rules.go
```diff
@@ -0,0 +1,64 @@
...
+				Type:     schema.TypeString,
+				Resolver: client.ResolveAWSRegion,
+			},
+			{
+				Name: "name",
+				Type: schema.TypeString,
+			},
+			{
+				Name: "rule_id",
+				Type: schema.TypeString,
```
roneli, line 36: A rule has more data, just like you did in GetWebAcl you need to do the same for Rules and RuleGroups

roneli (changes requested):
A few APIs missing to get more date (GetTagsForResource, GetRule, GetRuleGroup)

resources/waf_rules.go
michelvocks: @roneli Thanks! I was able to add tags for WebACLs but I was unable to get the tags for WafRules and WafRuleGroups because both resources don't return a resource arn and I wasn't able to get it at all.

resources/waf_rules.go
michelvocks: Thanks!

resources/waf_rules.go
michelvocks: After searching through the terraform aws code I was able to find a solution. Apparently it is possible to generate the ARN dynamically: 
https://github.com/hashicorp/terraform-provider-aws/blob/f4edafb9fed9d7112e6373adcbca152804213693/aws/resource_aws_waf_rule.go#L160-L166

roneli: If we are generating this column, it might be beneficial to add this a column to the Table as well.

michelvocks: Could you elaborate a little bit on what you mean with that? The `tags` column is defined in the `schema.Column` slice which should add it to the table, right?

roneli: I meant the "arn" column

roneli (approved):
Looks good, just respond to that last reviews comment about adding the "arn" column. (like you said in the todo, maybe we should create a generic helper to create an arn for any resource, in a different PR)

michelvocks:
> Looks good, just respond to that last reviews comment about adding the "arn" column. (like you said in the todo, maybe we should create a generic helper to create an arn for any resource, in a different PR)

Gotcha. I added a new column for the ARN. For now, I added a helper function that allows us to generate the ARN. Supposedly, we want to transform that helper function into a [client resolver](https://github.com/cloudquery/cq-provider-aws/blob/main/client/resolvers.go) but that requires us to add more information to the column struct (I guess) to be able to collect all the required information.

roneli:
> transform

Its a little more difficult to generate a column resolver, although if we add column dependencies between columns i.e ARN column requires region/resource_id etc' to be resolved first it allows this resolver to access them We can add this to another PR later on.

PR #90, opened 2021-06-05
Add schema documentation via tbls (without CI yet)

yevgenypats:
https://github.com/k1LoW/tbls

This will be integrated into the CI process
and checked that the latest doc is available via
`tbls diff`

Also, this will be used in cq-hub


roneli:
Duplicated from #95 which creates docs via SDK generator rather than tbls

PR #96, opened 2021-06-15
Updated the .goreleaser.yaml with arm64 build

madhuakula:
Currently, the cloudquery AWS provider doesn't have arm64 build in `.goreleaser.yaml`. Added `arm64` build for cq-provider-aws.


yevgenypats:
Closed with https://github.com/cloudquery/cq-provider-aws/pull/99

Issue #104, opened 2021-06-17
Add ElasticSearch resource


irmatov:
Add the ElasticSearch service based on user request cloudquery/cloudquery#125

PR #105, opened 2021-06-18
Add ElasticSearch resource

irmatov:


client/services.go
```diff
@@ -2,6 +2,7 @@ package client
 
 // define interfaces so we can easily mock AWS calls
 //go:generate mockgen -destination=./mocks/services.go -package=mocks . AutoscalingClient,ApigatewayClient,Apigatewayv2Client,CloudfrontClient,CloudtrailClient,CloudwatchClient,CloudwatchLogsClient,DirectconnectClient,Ec2Client,EcrClient,EfsClient,ElasticbeanstalkClient,ElbV1Client,ElbV2Client,EmrClient,FsxClient,IamClient,KmsClient,MQClient,OrganizationsClient,RdsClient,S3Client,SnsClient,EksClient,RedshiftClient,Route53Client,EcsClient,S3ManagerClient,LambdaClient,ConfigServiceClient,AnalyzerClient,WafClient,WafV2Client
+//go:generate mockgen -destination=./mocks/mock_elasticsearch.go -package=mocks . ElasticSearch
```
irmatov, line 5: still not sure if this is good idea. One drawback is that if we ever delete a resource (or rename destination file) it will require manually deleting mock_*.go files.

In other words, currently mock/services.go is regenerated always in one go. If we split go generate into multiple lines with multiple files we need to do rm mocks/mock_*.go before running `go generate` to ensure that all generated code is correct (no leftover unused/renamed files).

irmatov:
Closes #104

client/services.go
roneli: I agree with this we should change to your solution

roneli (approved):

PR #113, opened 2021-06-23
Test that all services are initialized

irmatov:
Uses reflection to inspect a struct returned by initServices
to ensure that there is no interface fields with nil values.

Example test failure (KMS initialization was removed):

    --- FAIL: Test_initServices_NoNilValues (0.00s)
        /home/thor/cloudquery/forks/cq-provider-aws/client/client_test.go:100: initServices().KMS == nil


michelvocks (approved):
Nice. Thanks for putting this together!

roneli (approved):
Nice! LGTM

PR #114, opened 2021-06-24
feat: Direct connect connections

James-Quigley:


roneli (approved):
Looks good, will merge soon. A little minor comment, will do this later, I would create a directconnect_connections_test.go instead of adding to the builders_test since that is deprecated. Its the same test more or less. (you can see other resources tables that have these new tests, we will update them all later on)

James-Quigley:
@roneli got it. Updated!

roneli (approved):

PR #115, opened 2021-07-04
PoC new SDK

roneli:


resources/ec2_instances.go
```diff
@@ -351,7 +352,7 @@ func Ec2Instances() *schema.Table {
 				Resolver:    fetchEc2InstanceBlockDeviceMappings,
 				Columns: []schema.Column{
 					{
-						Name:        "instance_id",
+						Name:        "instance_cq_id",
```
fdistorted, line 355: Should we add instance_id field that points to the instance id primary key of the parent?

roneli: yes, we can also do that with ParentFieldResolver

resources/iam_policies.go
```diff
@@ -89,11 +92,17 @@ func IamPolicies() *schema.Table {
 				Resolver:    fetchIamPolicyVersions,
 				Columns: []schema.Column{
```
fdistorted, line 93: Options are missing

resources/lambda_functions.go
```diff
@@ -678,6 +703,7 @@ func LambdaFunctions() *schema.Table {
 						Name:        "aws_lambda_function_version_layers",
 						Description: "An AWS Lambda layer (https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html). ",
 						Resolver:    fetchLambdaFunctionVersionLayers,
+						Options:     schema.TableCreationOptions{PrimaryKeys: []string{"function_version_id", "arn"}},
 						Columns: []schema.Column{
 							{
 								Name:        "function_version_id",
```
fdistorted, line 709: should be function_version_cq_id

resources/s3_buckets.go
```diff
@@ -21,6 +21,7 @@ func S3Buckets() *schema.Table {
 		IgnoreError:          client.IgnoreAccessDeniedServiceDisabled,
 		DeleteFilter:         client.DeleteAccountFilter,
 		PostResourceResolver: resolveS3BucketsAttributes,
+		Options:              schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "region", "name"}},
```
fdistorted, line 24: As I know buckets have global name across all the regions

resources/waf_rules.go
```diff
@@ -75,7 +78,7 @@ func WafRules() *schema.Table {
 					},
```
fdistorted, line 78: maybe  `data_id` should be  renamed  to `id`.

fdistorted (commented):
I found a few things to look at.
I think we need to set some rules to make naming more consistent and to have consistent priority of columns that can be used as pk

resources/cloudwatchlogs_filters.go
```diff
@@ -18,6 +18,7 @@ func CloudwatchlogsFilters() *schema.Table {
 		Multiplex:    client.AccountRegionMultiplex,
 		IgnoreError:  client.IgnoreAccessDeniedServiceDisabled,
 		DeleteFilter: client.DeleteAccountRegionFilter,
+		Options:      schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "region", "name", "log_group_name"}},
```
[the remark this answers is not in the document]
roneli: Not required since rc5.

resources/waf_rules.go
roneli: In relations I opted to not change. We can discuss this further

PR #122, opened 2021-07-22
feat: directconnect lags

James-Quigley:
Lag schema: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/directconnect@v1.4.1/types#Lag

As we are already capturing DirectConnect Connections in a separate table, I opted to not resave that data, and instead just save the connectionIds in a StringArray.

Connection -> Lag is a Many:1 relationship, but not all connections necessarily have a Lag. Perhaps this is a scenario where we could introduce nullable FKs or something like that?

### DIFF
--- a/resources/directconnect_lags.go
+++ b/resources/directconnect_lags.go
@@ -1,0 +1,213 @@
+package resources
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/service/directconnect"
+	"github.com/aws/aws-sdk-go-v2/service/directconnect/types"
+	"github.com/cloudquery/cq-provider-aws/client"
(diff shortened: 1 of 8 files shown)


resources/directconnect_lags.go
```diff
@@ -0,0 +1,213 @@
...
+				Resolver:    schema.PathResolver("LagId"),
+			},
+			{
+				Name:        "name",
+				Description: "The name of the LAG.",
+				Type:        schema.TypeString,
+				Resolver:    schema.PathResolver("LagName"),
+			},
+			{
+				Name:        "lag_state",
```
roneli, line 84: Rename to state?

roneli (changes requested):
Looks good, minor field rename and we can merge. I will release this still in beta

roneli (approved):

PR #132, opened 2021-08-03
fixed aws_iam_user tags resolver

zagronitay:


resources/iam_users.go
```diff
@@ -280,6 +281,7 @@ func fetchIamUsers(ctx context.Context, meta schema.ClientMeta, _ *schema.Resour
 			wUsers[i] = wrappedUser{
 				User:       u,
 				reportUser: ru,
+				isRoot:     false,
```
roneli, line 284: would also add this as a column to the aws_iam_user table

roneli (changes requested):
looks good, minor column add request

roneli (approved):
LGTM

PR #153, opened 2021-08-17
Use credentials cache

irmatov:
According to note here (https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#specify-credentials-programmatically) all providers must be wrapped in credentials cache.


irmatov:
Closes cloudquery/cloudquery-issues#88

PR #158, opened 2021-08-19
Fix multiplex bugs add migrations accordingly

roneli:


yevgenypats (approved):

resources/cognito_user_pools.go
```diff
@@ -474,7 +474,6 @@ func CognitoUserPools() *schema.Table {
 				Name:        "aws_cognito_user_pool_identity_providers",
 				Description: "A container for information about an identity provider.",
 				Resolver:    fetchCognitoUserPoolIdentityProviders,
-				Multiplex:   client.AccountRegionMultiplex,
```
fdistorted, line 477: shouldn't it have any multiplex?

fdistorted (approved):
looks good.
one small question

resources/cognito_user_pools.go
roneli: no, its a relation, we can't multiplex on multiplex it causes issues since you fetched a resource from region A and then you multiplex it's relation on all regions, bound to cause issues.

PR #177, opened 2021-09-07
Add ELBv2 attributes

irmatov:


irmatov:
Closes cloudquery/cloudquery#187

roneli (changes requested):
quick look I think attributes that we know exist should be transformed into columns rather than a json. 

see: 
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#load-balancer-attributes

https://docs.aws.amazon.com/elasticloadbalancing/latest/network/network-load-balancers.html#load-balancer-attributes

https://docs.aws.amazon.com/elasticloadbalancing/latest/gateway/gateway-load-balancers.html#load-balancer-attributes

resources/mq_brokers.go
```diff
@@ -180,7 +180,6 @@ func MqBrokers() *schema.Table {
 			{
 				Name:        "aws_mq_broker_configurations",
 				Resolver:    fetchMqBrokerConfigurations,
-				Multiplex:   client.AccountRegionMultiplex,
```
roneli, line 183: why did you remove the multiplex ?

resources/elbv2_load_balancers.go
```diff
@@ -219,3 +300,47 @@ func fetchElbv2LoadBalancerAvailabilityZoneAddresses(ctx context.Context, meta s
 	res <- p.LoadBalancerAddresses
 	return nil
 }
+
+type lbAttributes struct {
```
roneli, line 304: are there any "remain" attributes ?

roneli (changes requested):
Looks good. Left a few comments.

resources/mq_brokers.go
irmatov: Apparently, it is a multiplex on a dependent table, which doesn't make sense (and it even prints a warning, that is why I noticed and removed it).

resources/elbv2_load_balancers.go
irmatov: According to the documentation there shouldn't be any.

resources/mq_brokers.go
roneli: Good catch

resources/elbv2_load_balancers.go
roneli: okay

roneli (approved):

PR #192, opened 2021-09-13
fix byoip cidr errors

### LINKED ISSUE #190: BUG: resource: aws_ec2_byoip_cidrs. Error: The action DescribeByoipCidrs is not valid
The resource `aws_ec2_byoip_cidrs` Currently doesn't work and always return the following error:
```
1:43PM ERR received partial fetch error alias=aws err="table resolve error: operation error EC2: DescribeByoipCidrs, https response error StatusCode: 400, RequestID: 84580972-d912-4940-a849-2c07799f2324, api error InvalidAction: The action DescribeByoipCidrs is not valid for this web service." provider=aws table=aws_ec2_byoip_cidrs version=v0.5.13
```

roneli:
Closes #190

### DIFF
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -25,6 +25,19 @@ func IgnoreAccessDeniedServiceDisabled(err error) bool {
 	return false
 }
 
+func IgnoreWithInvalidAction(err error) bool {
+	if IgnoreAccessDeniedServiceDisabled(err) {
+		return true
+	}
+	var ae smithy.APIError
+	if errors.As(err, &ae) {
+		if ae.ErrorCode() == "InvalidAction" {
+			return true
+		}
+	}
+	return false
+}
+
 // GenerateResourceARN generates the arn for a resource.
 // Service: The service name e.g. waf or elb or s3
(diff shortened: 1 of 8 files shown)


client/helpers.go
```diff
@@ -25,6 +25,19 @@ func IgnoreAccessDeniedServiceDisabled(err error) bool {
 	return false
 }
 
+func IgnoreWithInvalidAction(err error) bool {
+	if IgnoreAccessDeniedServiceDisabled(err) {
+		return true
+	}
+	var ae smithy.APIError
+	if errors.As(err, &ae) {
+		if ae.ErrorCode() == "InvalidAction" {
```
yevgenypats, line 34: Why we need to ignore InvalidAction? Looks like something is wrong with our API or how we call it?

yevgenypats (changes requested):
Im not sure we need to ignore InvalidAction? why we are getting this error at all?

client/helpers.go
roneli: It only occurs on specific regions eu-west-3 etc' (only 4 regions failed) I doubled checked with CLI returns same result

PR #205, opened 2021-10-05
SQS resource

### LINKED ISSUE #202: Resource data for AWS SQS
We would like to run security audits of AWS SQS queues, but the resource data is currently missing.  Eg. queue is public due to resource policy, queue doesn't have encryption-at-rest, etc.

irmatov:
Closes #202


client/client.go
```diff
@@ -322,13 +324,17 @@ func initServices(region string, c aws.Config) Services {
 	awsCfg := c.Copy()
 	awsCfg.Region = region
 	return Services{
```
irmatov, line 326: this is just sorted & SQS is added

resources/provider.go
```diff
@@ -20,94 +20,95 @@ func Provider() *provider.Provider {
 		Migrations: awsMigrations,
 		ResourceMap: map[string]*schema.Table{
```
irmatov, line 21: this is sorted & sqs.queues is added

roneli (approved):
LGTM

PR #231, opened 2021-10-28
Lambda container e2e infra files

amanenk:
this pr contains infrastructural files to check containerized lambda functions
It does not work yet because of terraform docker provider bug


amanenk:
reopened because it has not only e2e files but logical fix

amanenk:
Logical fix mentioned above was found in another PR

PR #250, opened 2021-11-09
Implement Route53 Domains

irmatov:


resources/route53_domains.go
```diff
@@ -0,0 +1,552 @@
...
+			{
+				Name:        "tags",
+				Description: "A list of tags",
+				Type:        schema.TypeJSON,
+				Resolver:    resolveRoute53DomainTags,
+			},
+		},
+		Relations: []*schema.Table{
+			{
+				Name:        "aws_route53_domain_admin_contact_extra_params",
```
roneli, line 361: this relation can be a json, where name is the key and value is the value

resources/route53_domains.go
```diff
@@ -0,0 +1,552 @@
...
+					},
+					{
+						Name:        "glue_ips",
+						Description: "Glue IP address of a name server entry",
+						Type:        schema.TypeStringArray,
+					},
+				},
+			},
+			{
+				Name:        "aws_route53_domain_registrant_contact_extra_params",
```
roneli, line 407: same as admin extra params

resources/route53_domains.go
```diff
@@ -0,0 +1,552 @@
...
+					},
+					{
+						Name:        "value",
+						Description: "The value that corresponds with the name of an extra parameter.  This member is required.",
+						Type:        schema.TypeString,
+					},
+				},
+			},
+			{
+				Name:        "aws_route53_domain_tech_contact_extra_params",
```
roneli, line 430: same as admin extra params

roneli (changes requested):
Looks Good, A few minor changes required, would also add e2e tests if possible + update changelog.

resources/route53_domains.go
irmatov: Done

resources/route53_domains.go
irmatov: Done

resources/route53_domains.go
irmatov: Done

irmatov:
Changelog updated. Currently I don't see a way to provision this resource using Terraform. I suppose this resource actually registers the domain, so it's hardly testable.

resources/route53_domains.go
```diff
@@ -0,0 +1,492 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/route53domains"
+	"github.com/aws/aws-sdk-go-v2/service/route53domains/types"
+	"github.com/cloudquery/cq-provider-aws/client"
+	"github.com/cloudquery/cq-provider-sdk/provider/schema"
+)
+
+func Route53Domains() *schema.Table {
+	return &schema.Table{
```
roneli, line 16: Missing ignore error

resources/route53_domains.go
```diff
@@ -0,0 +1,492 @@
...
+		Name:         "aws_route53_domains",
+		Description:  "The domain names registered with Amazon Route 53.",
+		Resolver:     fetchRoute53Domains,
+		Multiplex:    client.AccountRegionMultiplex,
```
roneli, line 20: If I am not incorrect only us-east-1 works on route53domains service. https://github.com/aws/aws-cli/issues/1354 and see https://docs.aws.amazon.com/general/latest/gr/r53.html

roneli (changes requested):
Found minor issues in route53

resources/route53_domains.go
irmatov: Added:

Also added PK on this and relation table

resources/route53_domains.go
irmatov: Good catch, thanks!

Updated to use only the default region.

roneli (approved):
LGTM

PR #267, opened 2021-11-18
ecs clusters dependencies added

### LINKED ISSUE #260: AWS ECS Services
### New Resource

We already have ECS clusters but still miss information about what's running in them that's relevant for security.

* https://docs.aws.amazon.com/cli/latest/reference/ecs/describe-container-instances.html
* https://docs.aws.amazon.com/cli/latest/reference/ecs/describe-services.html

### Use Case

* ECS Service should not have tasks with privileged IAM access to an EC2 instance
* ECS Service tasks should not have access to EC2 instance metadata
* ECS Services should not have public IP addresses assigned to them
* ECS Service task role with admin privileges (should not have public IP addresses assigned to them)
* ECS Service should use a container image hosted on Amazon ECR
* ECS Service should expose only secure protocols on port 443
* ECS Cluster should not have running Container Instances with unconnected agents

amanenk:
fixes #260


roneli (approved):
LGTM, I do suggest adding e2e tests for these new dependencies

Issue #277, opened 2021-11-20
Store ARN of Security Groups


bbernays:
### New Resource
Store ARN of Security group in DB rather than relying on queries to generate it

### Use Case
Reduce complexity of queries by storing the ARN

PR #278, opened 2021-11-20
Add in Security Group ARN

bbernays:
fixes: #277

### DIFF
--- a/resources/config_configuration_recorders.go
+++ b/resources/config_configuration_recorders.go
@@ -90,8 +90,9 @@ func fetchConfigConfigurationRecorders(ctx context.Context, meta schema.ClientMe
 func generateConfigRecorderArn(_ context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
 	cl := meta.(*client.Client)
 	cfg, ok := resource.Item.(types.ConfigurationRecorder)
-	if ok {
-		return resource.Set(c.Name, client.GenerateResourceARN("config", "config-recorder", *cfg.Name, cl.Region, cl.AccountID))
+	if !ok {
+		return fmt.Errorf("not config config recorder")
 	}
-	return fmt.Errorf("not config config recorder")
+	return resource.Set(c.Name, client.GenerateResourceARN("config", "config-recorder", *cfg.Name, cl.Region, cl.AccountID))
+
 }
--- a/resources/config_configuration_recorders.go
+++ b/resources/config_configuration_recorders.go
@@ -2,6 +2,7 @@ package resources
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/aws/aws-sdk-go-v2/service/configservice/types"
 
@@ -88,6 +89,9 @@ func fetchConfigConfigurationRecorders(ctx context.Context, meta schema.ClientMe
 
 func generateConfigRecorderArn(_ context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
 	cl := meta.(*client.Client)
-	cfg := resource.Item.(types.ConfigurationRecorder)
-	return resource.Set(c.Name, client.GenerateResourceARN("config", "config-recorder", *cfg.Name, cl.Region, cl.AccountID))
+	cfg, ok := resource.Item.(types.ConfigurationRecorder)
+	if ok {
+		return resource.Set(c.Name, client.GenerateResourceARN("config", "config-recorder", *cfg.Name, cl.Region, cl.AccountID))
+	}
+	return fmt.Errorf("not config config recorder")
 }
--- a/resources/ec2_ebs_volumes.go
+++ b/resources/ec2_ebs_volumes.go
@@ -177,8 +177,8 @@ func fetchEc2EbsVolumeAttachments(_ context.Context, _ schema.ClientMeta, parent
 func resolveEbsVolumeArn(_ context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
 	cl := meta.(*client.Client)
 	ebs, ok := resource.Item.(types.Volume)
-	if ok {
-		return resource.Set(c.Name, client.GenerateResourceARN("ec2", "volume", *ebs.VolumeId, cl.Region, cl.AccountID))
+	if !ok {
+		return fmt.Errorf("not ec2 ebs volume")
 	}
-	return fmt.Errorf("not ec2 ebs volume")
+	return resource.Set(c.Name, client.GenerateResourceARN("ec2", "volume", *ebs.VolumeId, cl.Region, cl.AccountID))
 }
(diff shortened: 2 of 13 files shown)


resources/ec2_security_groups.go
```diff
@@ -475,3 +481,9 @@ func fetchEc2SecurityGroupIpPermissionsEgressUserIdGroupPairs(ctx context.Contex
 	res <- securityGroupIpPermissionEgress.UserIdGroupPairs
 	return nil
 }
+
+func resolveSGArn(_ context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+	cl := meta.(*client.Client)
+	sg := resource.Item.(types.SecurityGroup)
```
yevgenypats, line 487: Didn't touch that part of the code for a while but is a cast check needed here?

resources/ec2_security_groups.go
```diff
@@ -475,3 +481,9 @@ func fetchEc2SecurityGroupIpPermissionsEgressUserIdGroupPairs(ctx context.Contex
...
+	return resource.Set(c.Name, client.GenerateResourceARN("ec2", "security-group", *sg.GroupId, cl.Region, cl.AccountID))
```
yevgenypats, line 488: I think `if sg != nil` check is needed here.

yevgenypats (changes requested):
Looks good, had one comment.

But I think another review from @roneli is required here as I didn't touch this part of the code for a while.

resources/ec2_security_groups.go
```diff
@@ -475,3 +481,9 @@ func fetchEc2SecurityGroupIpPermissionsEgressUserIdGroupPairs(ctx context.Contex
 	res <- securityGroupIpPermissionEgress.UserIdGroupPairs
 	return nil
 }
+
+func resolveSGArn(_ context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
```
bbernays, line 485: This code comes from here:
https://github.com/cloudquery/cq-provider-aws/blob/c47fb629c96a50e960a036d516ce83352b0df79a/resources/ec2_ebs_volumes.go#L177-L181

resources/ec2_security_groups.go
roneli: yes, its advised, although unlikely to happen, its better than a panic.

resources/ec2_security_groups.go
roneli: no need since we just casted it

roneli (changes requested):
Missing migration + add to changelog under unreleased.

zagronitay (approved):
LGTM

resources/ec2_security_groups.go
disq: @roneli won't it still panic since it's not in the `sg, ok` syntax?

roneli: sorry, maybe my wording was incorrect. I suggest we do the `sg, ok` assertion syntax, since we prefer an error over a panic.

resources/ec2_security_groups.go
roneli: This should also do an assertion `ebs, ok` check.

resources/ec2_security_groups.go
bbernays: What error should I return?

resources/config_configuration_recorders.go
```diff
@@ -88,6 +89,9 @@ func fetchConfigConfigurationRecorders(ctx context.Context, meta schema.ClientMe
 
 func generateConfigRecorderArn(_ context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
 	cl := meta.(*client.Client)
-	cfg := resource.Item.(types.ConfigurationRecorder)
-	return resource.Set(c.Name, client.GenerateResourceARN("config", "config-recorder", *cfg.Name, cl.Region, cl.AccountID))
+	cfg, ok := resource.Item.(types.ConfigurationRecorder)
+	if ok {
```
disq, line 93: To make it Go-idiomatic the happy path needs to be the "unindented" one, so it should go like:

- run something
- check if error, { error handling block }
- continue processing

So this block would be:
```go
cfg, ok := resource.Item.(types.ConfigurationRecorder)
if !ok {
    return fmt.Errorf("not config config recorder")
}
return resource.Set(c.Name, client.GenerateResourceARN("config", "config-recorder", *cfg.Name, cl.Region, cl.AccountID))
```

Also in all `if ok` checks below... So, you should move error handling inside the if statement and the happy path outside.

disq (changes requested):

resources/ec2_ebs_volumes.go
```diff
@@ -176,6 +176,9 @@ func fetchEc2EbsVolumeAttachments(_ context.Context, _ schema.ClientMeta, parent
 
 func resolveEbsVolumeArn(_ context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
 	cl := meta.(*client.Client)
-	ebs := resource.Item.(types.Volume)
-	return resource.Set(c.Name, client.GenerateResourceARN("ec2", "volume", *ebs.VolumeId, cl.Region, cl.AccountID))
+	ebs, ok := resource.Item.(types.Volume)
+	if ok {
+		return resource.Set(c.Name, client.GenerateResourceARN("ec2", "volume", *ebs.VolumeId, cl.Region, cl.AccountID))
+	}
+	return fmt.Errorf("not ec2 ebs volume")
```
roneli, line 183: same comment that @disq gave on the check on sc

resources/route53_delegation_sets.go
```diff
@@ -69,6 +70,9 @@ func fetchRoute53DelegationSets(ctx context.Context, meta schema.ClientMeta, par
 	return nil
 }
 func resolveRoute53DelegationSetsArn(_ context.Context, _ schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
-	dl := resource.Item.(types.DelegationSet)
-	return resource.Set(c.Name, client.GenerateResourceARN("route53", "delegationset", *dl.Id, "", ""))
+	dl, ok := resource.Item.(types.DelegationSet)
+	if ok {
+		return resource.Set(c.Name, client.GenerateResourceARN("route53", "delegationset", *dl.Id, "", ""))
+	}
+	return fmt.Errorf("not route53 delegation set")
```
roneli, line 77: same CR as @disq gave in sc

roneli (changes requested):
LGTM (the migrations) would change the order of the assert as @disq suggested its more idiomatic

resources/config_configuration_recorders.go
bbernays: Thank you for clear feedback!! I have applied the change

disq (approved):

roneli (approved):
LGTM

PR #280, opened 2021-11-22
Add DMS Replication Instance

spangenberg:


roneli (changes requested):
LGTM would add the following https://docs.aws.amazon.com/dms/latest/APIReference/API_ListTagsForResource.html for Tags. You can pass a list of ARNs see cloudtrail for an example working with these types of APIs to reduce call overhead

roneli (approved):
LGTM, add e2e tests and we can merge.

Issue #282, opened 2021-11-22
S3 Account Settings


bbernays:
### New Resource

S3 Account settings are missing. We currently only support bucket level resources.

**Client**: [`s3control`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3control)

**API Method**: [`GetPublicAccessBlock`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/s3control#Client.GetPublicAccessBlock)

PR #285, opened 2021-11-22
Support S3 Account Settings

bbernays:
fixes: #282


resources/s3_accounts.go
```diff
@@ -0,0 +1,102 @@
...
+	var accountConfig s3control.GetPublicAccessBlockInput
+	accountConfig.AccountId = aws.String(c.AccountID)
+	s3AccountConfig.ConfigExists = false
+	resp, err := svc.GetPublicAccessBlock(ctx, &accountConfig, func(options *s3control.Options) {
+		options.Region = c.Region
+	})
+	if err != nil {
+		// If we received any error other than NoSuchPublicAccessBlockConfiguration, we return and error
+		var ae smithy.APIError
+		if errors.As(err, &ae) && ae.ErrorCode() != "NoSuchPublicAccessBlockConfiguration" {
```
yevgenypats, line 79: what will happen if it is returning `NoSuchPublicAccessBlockConfiguration` will resp be `nil`?

yevgenypats (dismissed):
Looks great, I love the `ConfigExist` attribute, this will make queries much more simpler. 
Only mock and e2e tests are missing.

resources/s3_accounts.go
```diff
@@ -0,0 +1,102 @@
...
+		s3AccountConfig.BlockPublicAcls = resp.PublicAccessBlockConfiguration.BlockPublicAcls
+		s3AccountConfig.BlockPublicPolicy = resp.PublicAccessBlockConfiguration.BlockPublicPolicy
+		s3AccountConfig.IgnorePublicAcls = resp.PublicAccessBlockConfiguration.IgnorePublicAcls
+		s3AccountConfig.RestrictPublicBuckets = resp.PublicAccessBlockConfiguration.RestrictPublicBuckets
+	} else {
+		s3AccountConfig.BlockPublicAcls = false
+		s3AccountConfig.BlockPublicPolicy = false
+		s3AccountConfig.IgnorePublicAcls = false
+		s3AccountConfig.RestrictPublicBuckets = false
+	}
```
roneli, line 93:
```suggestion
	if resp != nil {
		  res <- s3AccountConfig{resp.PublicAccessBlockConfiguration, true}
		}
```

resources/s3_accounts.go
roneli: If err == nil, resp should be != nil

roneli (changes requested):
LGTM, had one suggestion + as @yevgenypats we should add e2e + mock tests

resources/s3_accounts.go
disq: Use the const `s3control.ErrCodeNoSuchPublicAccessBlockConfiguration` instead of string?

bbernays: @disq - That seems to have been removed from the s3control package

bbernays: @disc- I have updated the error handling to use the native types from within the s3control/types package

resources/s3_accounts.go
```diff
@@ -0,0 +1,93 @@
...
+		// If we received any error other than NoSuchPublicAccessBlockConfiguration, we return and error
+		var nspabc *s3controlTypes.NoSuchPublicAccessBlockConfiguration
+		if !errors.As(err, &nspabc) {
+			return err
+		}
+		res <- s3AccountConfig
+	} else {
+		res <- S3AccountConfig{*resp.PublicAccessBlockConfiguration, true}
+	}
+
```
roneli, line 86:
```suggestion
		if errors.As(err, &nspabc) {
		  res <- s3AccountConfig{nil, false}
		}
	}
	
	res <- S3AccountConfig{*resp.PublicAccessBlockConfiguration, true}

```

roneli (approved):
LGTM, added one suggestion to make the flow cleaner, I think you can also add account id to the initialization and remove initialization in lines 69-71.

Still missing mock tests for sanity + e2e tests I think no terraform is required, just a fetching the data and verifying values worked.

disq (approved):

PR #289, opened 2021-11-23
Codebuild

### LINKED ISSUE #270: codebuild
### Use Case
https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/codebuild#Client.BatchGetProjects

amanenk:
fixes #270


resources/codebuild_projects.go
```diff
@@ -0,0 +1,782 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/codebuild"
+	"github.com/aws/aws-sdk-go-v2/service/codebuild/types"
+	"github.com/cloudquery/cq-provider-aws/client"
+	"github.com/cloudquery/cq-provider-sdk/provider/schema"
+)
+
+func CodebuildProjects() *schema.Table {
+	return &schema.Table{
```
roneli, line 16: Missing PK

resources/codebuild_projects.go
```diff
@@ -0,0 +1,782 @@
...
+			{
+				Name:        "webhook_url",
+				Description: "The URL to the webhook.",
+				Type:        schema.TypeString,
+				Resolver:    schema.PathResolver("Webhook.Url"),
+			},
+		},
+		Relations: []*schema.Table{
+			{
+				Name:        "aws_codebuild_project_environment_variables",
```
roneli, line 469: maybe we shouldn't save this since it might hold sensitive info, we can hash the values. @yevgenypats  taughts?

roneli (changes requested):
LGTM, had a few comments.

resources/codebuild_projects.go
fdistorted: One of the controls is related to this table.

roneli: Ah fair enough we can keep it, I am thinking if we can just json it as well, since type is basically the value type correct? so there is no point to save that info in a jsonb in theory we can cast it in golang to a map[string]interface{} but I think type is redundant perhaps.

amanenk: done

amanenk: we need to check type in one of controls

bbernays: Why don't we just store a hash of the value? The control only checks the key, not the value

roneli (approved):
LGTM

Issue #281, opened 2021-11-22
Store Data Showing Account Password Policy Doesn't Exist


bbernays:
### New Resource
Update `aws_iam_password_policies` with new field indicating whether the resource exists or not on a per account basis

### Use Case

When determining compliance it is important to understand the difference between a record not existing in the database and the resource not existing in AWS

PR #290, opened 2021-11-23
IAM password policy Improvements

bbernays:
Implements #281


resources/migrations/7_v0.6.1.up.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "aws_iam_password_policies" ADD COLUMN IF EXISTS policy_exists boolean;
```
roneli, line 1: this should be in v0.6.5 i.e for next release, same below

roneli (changes requested):
LGTM, just wrong migration file names

resources/migrations/7_v0.6.1.up.sql
bbernays: fixed

Issue #263, opened 2021-11-16
AWS Regions Enabled


bbernays:
### New Resource

Now that [AWS Regions](https://aws.amazon.com/blogs/security/setting-permissions-to-enable-accounts-for-upcoming-aws-regions/) are being brought online in disabled states we should be saving the state of every region in our asset inventory

### Use Case

For compliance validation, we can skip all regions that are disabled rather than inferring lack of data to mean an unknown state

### Additional context

All information can be grabbed from the ec2 DescribeRegions endpoint:

```
aws ec2 describe-regions --all-regions
```

roneli:
Was added in v0.7.0

PR #293, opened 2021-11-23
support regions table

roneli:


resources/provider.go
```diff
@@ -20,6 +20,7 @@ func Provider() *provider.Provider {
 		ErrorClassifier: client.ErrorClassifier,
 		Migrations:      awsMigrations,
 		ResourceMap: map[string]*schema.Table{
+			"regions":                               Regions(),
```
bbernays, line 23: should this be `aws.regions`?

bbernays (commented):
LGTM. What about e2e tests?

resources/provider.go
roneli: ye I can name it that

bbernays (approved):
LGTM... Did you decide not to change the name of the resource from `regions` to `aws.regions`?

roneli:
Closed #263

PR #294, opened 2021-11-23
Improve Doc Generation

bbernays:
Ran into issues when changing a table name where the github checks kept failing because the old table file was never being deleted. By clearing out all of the files for each run this will avoid that issue


docs/docs.go
```diff
@@ -9,7 +9,13 @@ import (
 )
 
 func main() {
-	err := docs.GenerateDocs(resources.Provider(), "./docs")
+	outputPath := "./docs"
+	err := os.RemoveAll(outputPath)
+	if err != nil {
+		fmt.Printf("%+v", err)
```
yevgenypats, line 15: Shouldn't it exit with non nil if err ?

yevgenypats: I think the next err is also missing `os.exit(1)`

yevgenypats (commented):
looks good few minor comments

roneli:
Would add this feature to the SDK as well so all providers can have this feature.

docs/docs.go
bbernays: Agreed

bbernays:
> Would add this feature to the SDK as well so all providers can have this feature.

I originally thought to just add it to the SDK but was concerned about adding destructive actions... wasn't sure if the directory was always 100% dynamically generated or if there could be manual additions

roneli:
> > Would add this feature to the SDK as well so all providers can have this feature.
> 
> I originally thought to just add it to the SDK but was concerned about adding destructive actions... wasn't sure if the directory was always 100% dynamically generated or if there could be manual additions

We can add it as an optional flag `delete_old bool` to the doc generator. I see it useful for all other providers we support.

bbernays:
@roneli - Created an issue on the cq-provider-sdk to implement this functionality

docs/docs.go
```diff
@@ -2,14 +2,31 @@ package main
...
+			fmt.Fprintf(os.Stderr, "Failed to generate docs: %s\n", err)
+			os.Exit(1)
+		}
+
+	}
+
+	err = docs.GenerateDocs(resources.Provider(), outputPath)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Failed to generate docs: %s\n", err)
 	}
```
roneli, line 32:
```suggestion
	if err = docs.GenerateDocs(resources.Provider(), outputPath); err != nil {
		fmt.Fprintf(os.Stderr, "Failed to generate docs: %s\n", err)
	}
```

docs/docs.go
```diff
@@ -2,14 +2,31 @@ package main
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to generate docs: %s\n", err)
+		os.Exit(1)
+	}
+	for _, d := range dir {
+		err = os.RemoveAll(path.Join([]string{outputPath + "/tables", d.Name()}...))
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to generate docs: %s\n", err)
+			os.Exit(1)
+		}
```
roneli, line 25:
```suggestion
		if err := os.RemoveAll(path.Join([]string{outputPath + "/tables", d.Name()}...)); err != nil {
			fmt.Fprintf(os.Stderr, "Failed to generate docs: %s\n", err)
			os.Exit(1)
		}
```

roneli (approved):
LGTM, added two suggestions

PR #300, opened 2021-11-25
feat/rename_dc_columns

roneli:
Fixed bad column names in directconnect, add missing cq_id columns for FK.


amanenk (changes requested):
I think we should rename direct_connect_gateway_id and direct_connect_gateway_name to id and name

amanenk (approved):

PR #301, opened 2021-11-25
Enhance config configuratiguration recorder with status

spangenberg:
Enhance for https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-pci-controls.html#pcidss-config-1


client/mocks/builders_test.go
```diff
@@ -224,6 +226,30 @@ func buildCloudwatchLogsFiltersMock(t *testing.T, ctrl *gomock.Controller) clien
 	}
 }
 
+func buildConfigConfigurationRecordersMock(t *testing.T, ctrl *gomock.Controller) client.Services {
```
roneli, line 229: I would move this to a new file, this is legacy testing code and we should slowly remove from it and depercate.

roneli (changes requested):
LGTM, would move the tests to a separate file, see comment in review.

resources/config_configuration_recorders.go
```diff
@@ -83,16 +120,57 @@ func fetchConfigConfigurationRecorders(ctx context.Context, meta schema.ClientMe
...
+	}
+	for _, configurationRecorder := range resp.ConfigurationRecorders {
+		var configurationRecorderStatus types.ConfigurationRecorderStatus
+		for _, s := range status.ConfigurationRecordersStatus {
+			if s.Name == configurationRecorder.Name {
+				configurationRecorderStatus = s
+				break
+			}
+		}
+		res <- configurationRecorderWrapper{
```
bbernays, line 144: Can we add an `exists` type field? That way we can have positive confirmation of the record in the database compared to having to handle if the record is missing for whatever reason (not in config, permissions issue, etc)

roneli (approved):
LGTM, can merge after fixing @bbernays suggestion

PR #315, opened 2021-11-30
Added logging configuration to WAF & WAFv2

zagronitay:


roneli (approved):
LGTM, tests should pass and we can merge

PR #318, opened 2021-12-01
Update EMR cluster with more info

irmatov:


irmatov:
This will require a migration. I'm not sure how to implement it. I'm inclined to do a "DROP TABLE aws_emr_cluster" during *UP* and do nothing on *DOWN*.

What is the right way?

roneli:
> This will require a migration. I'm not sure how to implement it. I'm inclined to do a "DROP TABLE aws_emr_cluster" during _UP_ and do nothing on _DOWN_.
> 
> What is the right way?

If you drop the table, than the table creator needs to create it on run if the user skips the build tables the resource won't work. You can do a drop -> create on the migration (you can extract the create table script from pgadmin for example)

irmatov:
Added migrations

roneli (approved):
LGTM

PR #321, opened 2021-12-01
Added secretsmanager secrets

zagronitay:


roneli (approved):
LGTM

resources/secretsmanager_secrets.go
```diff
@@ -0,0 +1,247 @@
...
+		})
+		if err != nil {
+			return err
+		}
+
+		// get more details about the secret
+		for _, n := range response.SecretList {
+
+			cfg := secretsmanager.DescribeSecretInput{
+				SecretId: n.ARN,
```
bbernays, line 157: I think you are missing a call to `ListSecretVersionIds` to get all of the different versions

bbernays (changes requested):
Looks good, but we don't have all of the versioning information about individual secrets

PR #330, opened 2021-12-02
Create Makefile

bbernays:
Make it easier to develop the `cq-provider-aws` 

you can run a specific e2e test like this:

```bash
make testName=TestIntegrationElasticbeanstal e2e-test
```


Makefile
```diff
@@ -0,0 +1,53 @@
+export CQ_PROVIDER_DEBUG=1
+export CQ_REATTACH_PROVIDERS=.cq_reattach
+
+# install the latest version of CQ
+install-cq:
+	curl -L https://github.com/cloudquery/cloudquery/releases/latest/download/cloudquery_Darwin_x86_64 -o cloudquery
+	chmod +x ./cloudquery
+
+install-cq-mac:
```
yevgenypats, line 9: what's the difference between `install-cq`, `install-cq-mac` (twice). I think we can get the OS and the arch automatically and put it in an env?

Makefile
```diff
@@ -0,0 +1,53 @@
...
+	curl -L https://github.com/cloudquery/cloudquery/releases/latest/download/cloudquery_Darwin_x86_64 -o cloudquery
+	chmod +x ./cloudquery
+
+
+# build the cq aws provider
+build:
+	go build -o cq-provider-aws
+
+# build and run the cq aws provider
+run: build
```
yevgenypats, line 23: maybe just `go run main.go`

yevgenypats (changes requested):
looks good, few comments.

Makefile
disq: If we can't, how about mac-m1? Or how about this:

```
install-cq:
@echo Go to https://docs.cloudquery.io/install-instructions and do the thing there please
```

(ie. I'm questioning whether it's needed to have it here)

bbernays: I left `install-cq` target in there because that is what I use. I would love to support automatically detecting arch and os.

Makefile
bbernays: So I had it like this because originally I was passing in flags in the build stage and wanted to only have to maintain them in a single place

Makefile
bbernays: I added the instructions under this target. When we can support dynamic detection we can replace it with the automated install

disq (approved):

disq (approved):

spangenberg (approved):
LGTM

PR #336, opened 2021-12-07
cloudfront migration fix

amanenk:


roneli (changes requested):
0.8.0 was released move this to 0.8.1 and ill release an update

resources/migrations/9_v0.8.1.up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE IF EXISTS "aws_cloudfront_distribution_default_cache_behavior_lambda_functions"
+    RENAME TO "aws_cloudfront_distribution_default_cache_behavior_functions";
+
+
+INSERT INTO aws_cloudfront_distribution_default_cache_behavior_functions (cq_id, distribution_cq_id, event_type,
+                                                                          lambda_function_arn, include_body)
+SELECT cq_id, distribution_cq_id, event_type, lambda_function_arn, include_body
+FROM aws_cloudfront_distribution_default_behaviour_lambda_functions;
```
roneli, line 8: no need to insert again, you just did a rename.

roneli (approved):
LGTM, remove insert and merge.

resources/migrations/9_v0.8.1.up.sql
amanenk: I think we need it because we have next situation: table `A` was renamed to table `B` in migration but in code it was renamed to table `C` which name is more than 64 symbols(we need to add a unit test for it). Fix renames table `C` to table `D` and inserts from table `B` to table `D`

PR #337, opened 2021-12-07
Fix call to ListAccounts in a non-org user

### LINKED ISSUE #254: Trying to call ListAccounts in a non-org user
### Describe the Bug
Currently ListAccounts is getting called on all accounts without any filter. Which produces the following warning:

```
Access Severity: Warning
Summary: Organizations: ListAccounts - You don't have permissions to access this resource.
Remediation: You are not authorized to perform this operation. Check your IAM policies, and ensure that you are using the correct access keys.
```

### Expected Behavior
Not calling ListAccounts in non-org user (research if there is a way to test it first if we are in a non org-user)

### CloudQuery Version
Version: 0.15.10
Commit: 00e576804344a17360d3ad94f79f6b30a4118c78
Date: 2021-11-14 09:54:28.360668 +0200 IST m=+0.005966954

### Steps to Reproduce
```
cloudquery fetch # using assume role
```

spangenberg:
The error might not be IAM policy related and needs to be treated
independent from the rest.

Fixes #254

### DIFF
--- a/client/errors.go
+++ b/client/errors.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 
-	organizationstypes "github.com/aws/aws-sdk-go-v2/service/organizations/types"
 	"github.com/aws/smithy-go"
 
 	"github.com/cloudquery/cq-provider-sdk/provider/schema"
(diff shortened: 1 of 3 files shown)


roneli (approved):
LGTM

roneli:
Another option, before I merge this, we can return that error from the resolver keeping the error classifier clean and generic. Wdyt?

client/errors.go
```diff
@@ -5,41 +5,48 @@ import (
...
 )
 
 func ErrorClassifier(meta schema.ClientMeta, resourceName string, err error) []diag.Diagnostic {
 	client := meta.(*Client)
+	var ee *diag.ExecutionError
+	if errors.As(err, &ee) {
+		return []diag.Diagnostic{
+			ee,
+		}
+	}
```
roneli, line 20: No need for this its already checked in https://github.com/cloudquery/cq-provider-sdk/blob/be1cd34f3527c627c1c7d814c5f4c90d3979e0aa/provider/provider.go#L247

roneli (changes requested):

roneli (approved):
LGTM, tests pass and merge

PR #338, opened 2021-12-08
added lambda runtime table

zagronitay:


resources/lambda_functions.go
```diff
@@ -916,6 +916,19 @@ func LambdaFunctions() *schema.Table {
 					},
 				},
 			},
+			{
+				Name:        "aws_lambda_runtimes",
```
roneli, line 920: why not just embed as a text array? also what is the relational FK here? can each function have only 1 runtime? if so just save it has a text field as runtime? Basically you will call this resolve each time per function and it will all conflict each time.

roneli (changes requested):

roneli (approved):
LGTM

roneli:
Would also add a test to this so we have sanity it works, no need to mock, just call and see it not empty

PR #353, opened 2021-12-16
Improve Docs And Support For Contributors

bbernays:
Implements the first piece of : https://github.com/cloudquery/cloudquery-issues/issues/192


roneli (approved):
Well needed docs

PR #356, opened 2021-12-17
Ensure maximum table name length

disq:
could add it in `cq-provider-sdk/provider/testing` instead maybe...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-###### SDK Version: 0.5.6
+###### SDK Version: 0.5.7
 
 ### 💥 Breaking Changes
 * Renamed columns of `aws_sagemaker_model_containers` image_config_repository_auth_config_repository_credentials_provider_arn -> image_config_repository_auth_config_repo_creds_provider_arn [#356](https://github.com/cloudquery/cq-provider-aws/pull/356).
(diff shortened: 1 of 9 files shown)


irmatov (approved):

roneli (approved):

roneli:
Would suggest to add these checks in SDK will open a referencing issue, or at least verify why it didn't work there.

disq:
@roneli not sure how to do that and keep it sane (it's a test but hang on it's not a test that's meant to run on the SDK itself... gets a provider and generates a test fn?)

Another issue, since the renamed column `image_config_repository_auth_config_repository_credentials_provider_arn` didn't exist (can't exist unless people have custom-compiled postgres with higher-limit identifiers), is the migration really needed?

disq:
we could attempt to have a smarter the migration and try to alter column rename `image_config_repository_auth_config_repository_credentials_prov` (the cut-out column name to 63 chars) to the target column maybe?

roneli:
> we could attempt to have a smarter the migration and try to alter column rename `image_config_repository_auth_config_repository_credentials_prov` (the cut-out column name to 63 chars) to the target column maybe?

You can try to rename on both if one exists it will just work

roneli:
> @roneli not sure how to do that and keep it sane (it's a test but hang on it's not a test that's meant to run on the SDK itself... gets a provider and generates a test fn?)
> 
> Another issue, since the renamed column `image_config_repository_auth_config_repository_credentials_provider_arn` didn't exist (can't exist unless people have custom-compiled postgres with higher-limit identifiers), is the migration really needed?

The SDK has a feature on provider serve it will fail if table added in resource map are in valid, it does the check on runtime, there might be a bug there.

disq:
> > we could attempt to have a smarter the migration and try to alter column rename `image_config_repository_auth_config_repository_credentials_prov` (the cut-out column name to 63 chars) to the target column maybe?
> 
> You can try to rename on both if one exists it will just work

apparently no need as it will truncate the identifier in the ALTER as well:
`NOTICE:  identifier "image_config_repository_auth_config_repository_credentials_provider_arn" will be truncated to "image_config_repository_auth_config_repository_credentials_prov"`

disq:
SDK will need one more upgrade

disq:
Ready to review/merge @roneli

roneli (approved):

PR #359, opened 2021-12-20
Run Fetch Tests on Merge to Main

bbernays:
**_Closes_**: [cloudquery-issues/issues/192](https://github.com/cloudquery/cloudquery-issues/issues/192)

**_Output from acceptance testing:_**
This pr doesn't touch any of the provider code, it is only adding more test invocations

Notes:

The one downside of this pr is that the messages to slack say "nightly"... not sure if it is worth changing to make more generic


resources/ec2_security_groups.go
```diff
@@ -406,23 +390,43 @@ func fetchEc2SecurityGroupIpPermissions(ctx context.Context, meta schema.ClientM
...
-	if !ok {
-		return fmt.Errorf("not ec2 security group ip permission")
+
+	type customIpRange struct {
+		cidr        string
+		description string
+		cidr_type   string
+	}
+
+	ipRanges := make([]customIpRange, len(securityGroupIpPermission.IpRanges)+len(securityGroupIpPermission.Ipv6Ranges))
```
disq, line 421: maybe just use `make([]T, 0, capacity)` and use `append()` below to simplify the reading and avoid any future errors.

disq (approved):
LGTM, minor nit

resources/ec2_security_groups.go
bbernays: Sorry didn't mean to include this change in this pr. I have removed it, but will apply your changes to my local branch

disq (approved):

PR #361, opened 2021-12-21
fixed pk error - secretmanager

### LINKED ISSUE #354: Duplicate key on secret manager resource
### Describe the Bug
Secret manager resource has duplicate key inserted, ARN might not be unique enough, and we might require account_id/region. Double check if this is a per region resource or not.

### Expected Behavior
No duplicate keys

### CloudQuery Version
v0.17.3

zagronitay:
Relates OR Closes #354 

```
...
```

### DIFF
--- a/resources/secretsmanager_secrets.go
+++ b/resources/secretsmanager_secrets.go
@@ -150,9 +150,10 @@ func fetchSecretsmanagerSecrets(ctx context.Context, meta schema.ClientMeta, _ *
 			return err
 		}
 
+		var secrets []WrappedSecret
+
 		// get more details about the secret
 		for _, n := range response.SecretList {
-			secretEntry := n
 			cfg := secretsmanager.DescribeSecretInput{
 				SecretId: n.ARN,
 			}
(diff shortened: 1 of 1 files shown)


spangenberg (approved):
LGTM

resources/secretsmanager_secrets.go
```diff
@@ -164,7 +164,7 @@ func fetchSecretsmanagerSecrets(ctx context.Context, meta schema.ClientMeta, _ *
 			}
 
 			secret := WrappedSecret{
-				SecretListEntry:   &n,
```
disq, line 167: Not sure if it'll work as `range` re-uses the value. Adding a `n := n` before, or accessing the element with `response.SecretList[i]` would be safer.

disq (approved):

resources/secretsmanager_secrets.go
```diff
@@ -152,7 +152,7 @@ func fetchSecretsmanagerSecrets(ctx context.Context, meta schema.ClientMeta, _ *
 
 		// get more details about the secret
 		for _, n := range response.SecretList {
```
roneli, line 154: would create a list of secret entries and pass them as one to channel instead of one by one, that would have also avoided this PK issue, you can keep this fix as well.

roneli (changes requested):

roneli (approved):

PR #364, opened 2021-12-22
Bugfix/ingore dns error

zagronitay:
Relates OR Closes #0000

Output from acceptance testing:
More information about running the tests [here](../docs/contributing/e2e_tests.md)

```
$ make testName=TestAccXXX e2e-test-with-apply
...
```

PR #366, opened 2021-12-22
Revert "Bugfix/ingore dns error"

yevgenypats:
Reverts cloudquery/cq-provider-aws#364


yevgenypats:
Reverted as it doesn't solve it just an internal ip that is different in every network/computer

zagronitay (approved):
LGTM

PR #367, opened 2021-12-22
fix region filtering

yevgenypats:
OptInStatus can be `not-opted-in` which means the region is disabled.

Relates OR Closes #0000

```
...
```


amanenk (approved):
LGTM

roneli (approved):

PR #368, opened 2021-12-22
DescribeConformancePacks always returns AccessDeniedException in af-south-1

yevgenypats:
This is a workaround untill this bug is fixed
https://github.com/aws/aws-sdk-go-v2/issues/1539

Relates OR Closes #0000

```
...
```

### DIFF
--- a/resources/config_conformance_pack.go
+++ b/resources/config_conformance_pack.go
@@ -94,7 +94,7 @@ func fetchConfigConformancePacks(ctx context.Context, meta schema.ClientMeta, pa
 			options.Region = c.Region
 		})
 
-		// This is a workaround untill this bug is fixed = https://github.com/aws/aws-sdk-go-v2/issues/1539
+		// This is a workaround until this bug is fixed = https://github.com/aws/aws-sdk-go-v2/issues/1539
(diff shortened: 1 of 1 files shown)


resources/config_conformance_pack.go
```diff
@@ -85,10 +88,17 @@ func ConfigConformancePack() *schema.Table {
...
 	c := meta.(*client.Client)
 	config := configservice.DescribeConformancePacksInput{}
+	var ae smithy.APIError
 	for {
 		resp, err := c.Services().ConfigService.DescribeConformancePacks(ctx, &config, func(options *configservice.Options) {
 			options.Region = c.Region
 		})
+
+		// This is a workaround untill this bug is fixed = https://github.com/aws/aws-sdk-go-v2/issues/1539
+		if c.Region == "af-south-1" && errors.As(err, &ae) && ae.ErrorCode() == "AccessDeniedException" {
```
amanenk, line 98: Does it happen only in af-south region? Should we enable all the regions and test it.

yevgenypats: I enabled all of them and it happens only in af-south-1 even if it enabled. this is why I opened the issue in aws github

amanenk (approved):

PR #369, opened 2021-12-22
ignore AccessDeniedException completely

yevgenypats:
I suggest ignoring this completely until AWS will come up with a better way to detect if we are in a root account or not, otherwise it makes the following two issues (I know this was the original suggestion of @spangenberg but I missed this PR or didn't understand completely how the diag works ):

1) the user experience on correct permissions is not smooth
2) our ci tests might fail because of that

Relates OR Closes #0000

```
...
```

### DIFF
--- a/resources/organizations_accounts.go
+++ b/resources/organizations_accounts.go
@@ -84,7 +84,7 @@ func fetchOrganizationsAccounts(ctx context.Context, meta schema.ClientMeta, par
 			case "AWSOrganizationsNotInUseException", "AccessDeniedException":
 				// This is going to happen most probably due to account not being the root organizational account
 				// so it's better to ignore it completly as it happens basically on every account
-				// otherwise it screws up with dev experiance and with our tests
+				// otherwise it screws up with dev experience and with our tests
 				meta.Logger().Warn("account is probably not the root organization account https://docs.aws.amazon.com/organizations/latest/APIReference/API_ListAccounts.html")
 				return nil
 			}
--- a/resources/organizations_accounts.go
+++ b/resources/organizations_accounts.go
@@ -11,7 +11,6 @@ import (
 	"github.com/cloudquery/cq-provider-aws/client"
 
 	"github.com/cloudquery/cq-provider-sdk/provider/schema"
-	"github.com/cloudquery/cq-provider-sdk/provider/schema/diag"
 )
 
 func OrganizationsAccounts() *schema.Table {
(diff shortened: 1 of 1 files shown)


spangenberg (approved):
LGTM 👍🏻

resources/organizations_accounts.go
```diff
@@ -83,7 +83,11 @@ func fetchOrganizationsAccounts(ctx context.Context, meta schema.ClientMeta, par
 		if errors.As(err, &ae) {
 			switch ae.ErrorCode() {
 			case "AWSOrganizationsNotInUseException", "AccessDeniedException":
-				return diag.FromError(err, diag.IGNORE, diag.ACCESS, OrganizationsAccounts().Name, client.ParseSummaryMessage(c.Accounts, err, ae), "Missing permissions or account might not be root/organizational unit.")
+				// This is going to happen most proabbly due to account not being the root organizational account
+				// so it's better to ignore it completly as it happens basically on every account
+				// otherwise it screws up with dev experiance and with our tests
+				log.Warn("account is most probabbly not the root organization account https://docs.aws.amazon.com/organizations/latest/APIReference/API_ListAccounts.html")
```
disq, line 89:
```suggestion
				log.Warn("account is most probabbly not the root organization account https://docs.aws.amazon.com/organizations/latest/APIReference/API_ListAccounts.html")
```
```suggestion
				log.Warn("account is most probably not the root organization account https://docs.aws.amazon.com/organizations/latest/APIReference/API_ListAccounts.html")
```

(with the typo fixed the wording still isn't the best)

resources/organizations_accounts.go
```diff
@@ -83,7 +83,11 @@ func fetchOrganizationsAccounts(ctx context.Context, meta schema.ClientMeta, par
 		if errors.As(err, &ae) {
 			switch ae.ErrorCode() {
 			case "AWSOrganizationsNotInUseException", "AccessDeniedException":
-				return diag.FromError(err, diag.IGNORE, diag.ACCESS, OrganizationsAccounts().Name, client.ParseSummaryMessage(c.Accounts, err, ae), "Missing permissions or account might not be root/organizational unit.")
+				// This is going to happen most proabbly due to account not being the root organizational account
```
disq, line 86:
```suggestion
				// This is going to happen most probably due to account not being the root organizational account
```

resources/organizations_accounts.go
```diff
@@ -75,6 +74,7 @@ func OrganizationsAccounts() *schema.Table {
 // ====================================================================================================================
 func fetchOrganizationsAccounts(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan interface{}) error {
 	c := meta.(*client.Client)
+	log := meta.Logger()
```
roneli, line 77: would just do meta.Logger().Warn in line 89 and remove this variable

roneli (approved):

PR #370, opened 2021-12-22
Provider Version

bbernays:
Relates to [cloudquery-issues/issues/203](https://github.com/cloudquery/cloudquery-issues/issues/203)

This inserts the version of the provider into a place where the SDK can access it

I couldn't come up with a way to do this on the SDK side. If anyone has an idea feel free to let me know and I would be happy to attempt it!


amanenk (approved):
LGTM

roneli (approved):

PR #376, opened 2021-12-23
Add test: MultiplexerPKMatch

disq:
Not sure if this is even valid, only issue so far is elasticsearch (which can have the same `id` under multiple regions...) but this is failing so many  resources.


yevgenypats:
Can we do that without a test? i.e create cloudquery linter  as part of cq-provider-sdk?

disq:
> Can we do that without a test? i.e create cloudquery linter as part of cq-provider-sdk?

I don't think it makes sense to make this a linter as it would limit the ways you could define table columns. Would be limiting in an SDK.

Also much harder to develop I think.

PR #378, opened 2021-12-24
feat: Add ARNs part 1

irmatov:
Add ARNs to following resources:

* aws_apigateway_api_keys
* aws_apigateway_client_certificates
* aws_apigateway_domain_names
* aws_apigateway_domain_name_base_path_mappings
* aws_apigateway_rest_apis
* aws_apigateway_rest_api_authorizers
* aws_apigateway_rest_api_deployments
* aws_apigateway_rest_api_documentation_parts
* aws_apigateway_rest_api_documentation_versions
* aws_apigateway_rest_api_gateway_responses
* aws_apigateway_rest_api_models
* aws_apigateway_rest_api_request_validators
* aws_apigateway_rest_api_resources
* aws_apigateway_rest_api_stages
* aws_apigateway_usage_plans
* aws_apigateway_usage_plan_keys
* aws_apigateway_vpc_links
* aws_apigatewayv2_apis
* aws_apigatewayv2_api_authorizers
* aws_apigatewayv2_api_deployments
* aws_apigatewayv2_api_integrations
* aws_apigatewayv2_api_integration_responses
* aws_apigatewayv2_api_models
* aws_apigatewayv2_api_routes
* aws_apigatewayv2_api_route_responses
* aws_apigatewayv2_api_stages


irmatov:
https://docs.aws.amazon.com/apigateway/latest/developerguide/arn-format-reference.html

client/helpers.go
```diff
@@ -149,3 +151,38 @@ func accountObfusactor(aa []Account, msg string) string {
...
+		}
+		return resource.Set(c.Name, arn.ARN{
+			Partition: "aws",
+			Service:   string(service),
+			Region:    cl.Region,
+			AccountID: accountID,
+			Resource:  strings.Join(append([]string{""}, idParts...), "/"),
+		}.String())
+	}
+}
```
bbernays, line 188: Is this new function completely necessary can some of the functionality from  `client.GenerateResourceARN` be used?

disq (approved):

client/helpers.go
irmatov: This changes a bit in follow up PR, and I think I will remove `client.GenerateResourceARN` completely at the end.

resources/services/apigateway/apigateway_domain_names.go
```diff
@@ -146,6 +160,22 @@ func ApigatewayDomainNames() *schema.Table {
 						Type:        schema.TypeUUID,
 						Resolver:    schema.ParentIdResolver,
 					},
+					{
+						Name:        "arn",
+						Description: "The Amazon Resource Name (ARN) for the resource.",
+						Type:        schema.TypeString,
+						Resolver: client.ResolveARN(client.ApigatewayService, func(resource *schema.Resource) ([]string, error) {
+							r, ok := resource.Item.(types.BasePathMapping)
+							if !ok {
```
yevgenypats, line 169: I think this check is not necessary here as we check it in tests and it's not something user controlled.

yevgenypats (changes requested):
all type assertion checks are not necessary here as we check it in the tests and as it's not user controlled it's not possible this will crash in prod and not in tests. moreover as it's not user controlled the user wont have anything to do with such error.

I'll give it another pass after this.

resources/services/apigateway/apigateway_domain_names.go
irmatov: Removed

yevgenypats (approved):

PR #384, opened 2021-12-30
Fix elasticsearch domains PK

disq:
ElasticSearch domains [are unique](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/es/create-elasticsearch-domain.html) _per region_ but the PK was defined as `account_id, id`.

Manually tested in a "throwaway" AWS account.

### DIFF
--- a/resources/services/elasticsearch/elasticsearch_domains.go
+++ b/resources/services/elasticsearch/elasticsearch_domains.go
@@ -18,7 +18,7 @@ func ElasticsearchDomains() *schema.Table {
 		Multiplex:    client.ServiceAccountRegionMultiplexer("es"),
 		IgnoreError:  client.IgnoreAccessDeniedServiceDisabled,
 		DeleteFilter: client.DeleteAccountRegionFilter,
-		Options:      schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "id"}},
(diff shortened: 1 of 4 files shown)


resources/services/elasticsearch/elasticsearch_domains.go
```diff
@@ -18,7 +18,7 @@ func ElasticsearchDomains() *schema.Table {
 		Multiplex:    client.ServiceAccountRegionMultiplexer("es"),
 		IgnoreError:  client.IgnoreAccessDeniedServiceDisabled,
 		DeleteFilter: client.DeleteAccountRegionFilter,
-		Options:      schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "id"}},
+		Options:      schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "id", "region"}},
```
roneli, line 21:
```suggestion
		Options:      schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "region", "id"}},
```

This is a better PK and more inline with the rest of region defined pks, would require a change in the SQL.

roneli (changes requested):
LGTM, had minor nitpick on pk order

roneli (approved):

PR #391, opened 2022-01-05
Support Wildcard for Region

bbernays:
- [x] Relates OR Closes cloudquery-issues/issues/242

```

```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -311,7 +311,11 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 
 	for _, account := range awsConfig.Accounts {
 		// Use a copy of the full region list for each account
-		localRegions := client.regions
+		localRegions := account.Regions
+		if len(localRegions) == 0 {
+			localRegions = client.regions
+		}
+
 		var err error
 		var awsCfg aws.Config
 
@@ -390,19 +394,19 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 		if err != nil {
 			return nil, fmt.Errorf("failed to find disabled regions for account %s. AWS Error: %w", accountID, err)
 		}
-		localRegions = filterDisabledRegions(localRegions, res.Regions, wildcardAllRegions)
+		account.Regions = filterDisabledRegions(localRegions, res.Regions, wildcardAllRegions)
 
-		if len(localRegions) == 0 {
+		if len(account.Regions) == 0 {
 			return nil, fmt.Errorf("no enabled regions provided in config for account %s", accountID)
 		}
 
 		if client.AccountID == "" {
 			// set default
 			client.AccountID = *output.Account
-			client.Region = localRegions[0]
+			client.Region = account.Regions[0]
 			client.Accounts = append(client.Accounts, Account{ID: *output.Account, RoleARN: *output.Arn})
 		}
-		for _, region := range localRegions {
+		for _, region := range account.Regions {
 			client.ServicesManager.InitServicesForAccountAndRegion(*output.Account, region, initServices(region, awsCfg))
 		}
 	}
--- a/client/client.go
+++ b/client/client.go
@@ -291,6 +291,11 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 		client.regions = allRegions
 		logger.Info(fmt.Sprintf("No regions specified in config.yml. Assuming all %d regions", len(client.regions)))
 	}
+	var wildcardAllRegions bool
+	if len(client.regions) == 1 && client.regions[0] == "*" {
+		client.regions = allRegions
+		wildcardAllRegions = true
+	}
 
 	if len(awsConfig.Accounts) == 0 {
 		awsConfig.Accounts = append(awsConfig.Accounts, Account{
@@ -379,7 +384,7 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 		if err != nil {
 			return nil, fmt.Errorf("failed to find disabled regions for account %s. AWS Error: %w", accountID, err)
 		}
-		client.regions = filterDisabledRegions(client.regions, res.Regions)
+		client.regions = filterDisabledRegions(client.regions, res.Regions, wildcardAllRegions)
 
 		if len(client.regions) == 0 {
 			return nil, fmt.Errorf("no enabled regions provided in config for account %s", accountID)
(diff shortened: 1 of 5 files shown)


client/client.go
```diff
@@ -473,9 +478,15 @@ func filterDisabledRegions(regions []string, enabledRegions []types.Region) []st
 	}
 
 	var filteredRegions []string
-	for _, r := range regions {
-		if regionsMap[r] {
-			filteredRegions = append(filteredRegions, r)
+	if allRegions {
+		for region := range regionsMap {
+			filteredRegions = append(filteredRegions, region)
+		}
```
bbernays, line 484: This prevents there being regions not in our list, but that get returned by `ec2.DescribeRegions`

disq (approved):
LGTM

client/client.go
yevgenypats: Interesting edge-case, I'd add a comment in the code

yevgenypats (changes requested):
Looks like when no regions are specified and `["*"]` it should be the same behaviour, I would merge between the two logics.

yevgenypats:
While reviewing found an unrelated but to this PR but related to this code section - https://github.com/cloudquery/cq-provider-aws/issues/392

bbernays:
> Looks like when no regions are specified and `["*"]` it should be the same behaviour, I would merge between the two logics.

Yeah I can do that. Originally left them separate because I added the more complete logic for handling all regions, in order to maintain backwards compatibility... But I have merged the 2 now

client/client.go
```diff
@@ -287,9 +287,18 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 	awsConfig := providerConfig.(*Config)
 	client := NewAwsClient(logger, awsConfig.Accounts, awsConfig.Regions)
 
-	if len(client.regions) == 0 {
+	// validate regions values
+	for i, region := range client.regions {
+		if i != 0 && region == "*" {
+			return nil, fmt.Errorf("region wildcard \"*\" is only supported in 0 index")
```
disq, line 293:
```suggestion
			return nil, fmt.Errorf("region wildcard \"*\" is only supported as first argument")
```
(Let's make it user friendly)

client/client.go
```diff
@@ -282,15 +250,27 @@ func (c *Client) withAccountIDRegionAndNamespace(accountID, region, namespace st
...
+func validateRegions(logger hclog.Logger, regions []string) (bool, error) {
+	// validate regions values
+	for i, region := range regions {
+		if i != 0 && region == "*" {
+			return false, fmt.Errorf("region wildcard \"*\" is only supported as first argument")
+		}
+	}
+
+	var wildcardAllRegions bool
+	if len(regions) == 1 && regions[0] == "*" || len(regions) == 0 {
```
yevgenypats, line 262: I think it's better to have `()` around the first check just to be on the safe side

client/client.go
```diff
@@ -282,15 +250,27 @@ func (c *Client) withAccountIDRegionAndNamespace(accountID, region, namespace st
 	}
 }
 
+func validateRegions(logger hclog.Logger, regions []string) (bool, error) {
```
yevgenypats, line 253: I'd add documentation on this method as in additional to `validateRegions` it return if it's also wildcard.

client/client.go
```diff
@@ -282,15 +250,27 @@ func (c *Client) withAccountIDRegionAndNamespace(accountID, region, namespace st
...
+	// validate regions values
+	for i, region := range regions {
+		if i != 0 && region == "*" {
+			return false, fmt.Errorf("region wildcard \"*\" is only supported as first argument")
+		}
+	}
+
+	var wildcardAllRegions bool
+	if len(regions) == 1 && regions[0] == "*" || len(regions) == 0 {
+		logger.Info("All regions specified in config.yml. Assuming all regions")
```
yevgenypats, line 263: I think we can remove the dependency for the logger here and move this to an outside of this function. This way this function is more pure and can be more easily re-used in other-places.

client/client.go
yevgenypats: Another idea is to split it to two functions. `isValidRegions(regions []string) error` and `isAllRegions(regions []string) bool` this way it will be also more self-explanatory.

client/client.go
```diff
@@ -372,26 +362,26 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
...
-					o.Region = awsConfig.Regions[0]
+				if len(localRegions) > 0 && !wildcardAllRegions {
+					o.Region = localRegions[0]
 				}
 			})
 		if err != nil {
 			return nil, fmt.Errorf("failed to find disabled regions for account %s. AWS Error: %w", accountID, err)
 		}
-		client.regions = filterDisabledRegions(client.regions, res.Regions)
+		account.Regions = filterDisabledRegions(localRegions, res.Regions, wildcardAllRegions)
```
yevgenypats, line 372: My personal preference is to remove the need for `wildcardAllRegions` as this can be inferred from `localRegions` within `filterDisabledRegions` - this way it will remove any potential bugs and/or inconsistencies of `localRegions` and `wildcardAllRegions` variables

yevgenypats (changes requested):
Looks good, I think the logic is correct - a few minor suggestion regarding code structure.

client/client.go
```diff
@@ -287,9 +287,18 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
...
+	for i, region := range client.regions {
+		if i != 0 && region == "*" {
+			return nil, fmt.Errorf("region wildcard \"*\" is only supported in 0 index")
+		}
+	}
+
+	var wildcardAllRegions bool
+	if len(client.regions) == 1 && client.regions[0] == "*" || len(client.regions) == 0 {
+		logger.Info(fmt.Sprintf("All regions specified in config.yml. Assuming all %d regions", len(allRegions)))
 		client.regions = allRegions
```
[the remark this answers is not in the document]
bbernays: I think I might have already fixed this... let me know if you think it is still wrong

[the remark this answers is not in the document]
yevgenypats: Yes, ignore this  - old comment but got to this review somehow.

client/client.go
bbernays: fixed

client/client.go
bbernays: fixed

client/client.go
bbernays: changed

client/client.go
```diff
@@ -282,15 +250,28 @@ func (c *Client) withAccountIDRegionAndNamespace(accountID, region, namespace st
...
+	for i, region := range regions {
+		if i != 0 && region == "*" {
+			return fmt.Errorf("region wildcard \"*\" is only supported as first argument")
+		}
+	}
+	return nil
+}
+func isAllRegions(regions []string) bool {
+	wildcardAllRegions := false
+	if (len(regions) == 1) && (regions[0] == "*" || len(regions) == 0) {
```
yevgenypats, line 264: I think this should be `(len(regions) == 1 && regions[0] == "*") || len(regions) == 0`

yevgenypats (changes requested):
Looks good. One comment (I think a bug) and ready for merge.

client/client.go
bbernays: you are right. Caught it in the test cases

yevgenypats (approved):
Awesome!

PR #393, opened 2022-01-06
fix: Ignore NoSuchBucket error

yevgenypats:
This is something we are seeing in sentry right now. giving a try to fix it.

- [ ] Relates OR Closes #0000

```

```


resources/services/s3/buckets.go
```diff
@@ -472,6 +472,13 @@ func resolveS3BucketsAttributes(ctx context.Context, meta schema.ClientMeta, res
 		return err
 	}
 	if err := resolveBucketLogging(ctx, meta, resource, *r.Name, bucketRegion); err != nil {
+		if errors.As(err, &ae) && ae.ErrorCode() == "NoSuchBucket" {
+			// Not sure why happens here and not earlier in GetBucketRegion. Maybe it's being deleted
+			// while fetching. In that case we might need to add this to global ignore but would like to
+			// add this here first.
+			log.Debug("resolveBucketLogging: Skipping bucket (already deleted)", "bucket", *r.Name)
+			return nil
```
bbernays, line 480: I think this has to do with a regional issue of S3. S3 service I think has to be using the regional endpoint to get the BucketLogging configs.

PR #394, opened 2022-01-06
Client Filter Regions

### LINKED ISSUE #392: Regions config are global and overriding all accounts
### Describe the bug

Currently we have the ability to specify either all or subset of regions for our accounts. 

Currently the implementation of this logic is flawed - https://github.com/cloudquery/cq-provider-aws/blob/main/client/client.go#L382

As we can see if for example the last account have all regions disabled it will also disable the fetching for all regions in the first account.

### Expected Behavior

our `Client` should have Regions per Account (even if we currently give the ability to filter regions on a global basis).

### Steps to Reproduce

https://github.com/cloudquery/cq-provider-aws/blob/main/client/client.go#L382

### Possible Solution

_No response_

main

### Additional Context

_No response_

bbernays:
- [x] Relates OR Closes #392 
- [x] Added integration tests as described [here](https://docs.cloudquery.io/docs/developers/sdk/testing)

```

```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -299,6 +299,7 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 			RoleARN:   defaultVar,
 		})
 	}
+	// Use a copy of the full region list
 	localRegions := client.regions
 	for _, account := range awsConfig.Accounts {
 		var err error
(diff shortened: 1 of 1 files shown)


spangenberg (approved):

yevgenypats (changes requested):
I would merge between this PR and this https://github.com/cloudquery/cq-provider-aws/pull/391 as I think they would conflict otherwise (merge I mean just use one of them and close the other)

bbernays:
feature in pr #391

PR #395, opened 2022-01-08
WIP: Fix regions filtering overriding for each account

### LINKED ISSUE #242: Disabled regions detection is using hardcoded region "us-east-1" that could be a disabled region in some configurations
### Describe the Bug
In some configurations, accounts could be regional and thus could not use the region "us-east-1", disabled by Service control policies (SCPs).

screenshot_318

### Expected Behavior
If regions are specified in the config.hcl, the use of _DescribeRegions_ to detect disable regions should use one of them. 

### CloudQuery Version
cq-provider-aws v0.6.0

### Debug Output

cloudquery-AWS-disabledRegion

### Steps to Reproduce
Use of the following configuration with an account limited to these regions :
```
provider "aws" {
  configuration {
    // Optional. if you want to assume role to multiple account and fetch data from them
    //accounts "<YOUR ID>" {
    // Optional. Role ARN we want to assume when accessing this account
    // role_arn = <YOUR_ROLE_ARN>
    // }
    regions = ["eu-west-3", "eu-central-1"]
    // Optional. Enable AWS SDK debug logging.
    aws_debug = true
  }
```

### Additionnal content
I will probably do a PR for this one. I might have the time for it.

### References
https://github.com/cloudquery/cq-provider-aws/issues/39

yevgenypats:
- [ ] Closes #242
- [ ] Related #391 

This is a continuation of https://github.com/cloudquery/cq-provider-aws/pull/391. 

As uncovered in #391 this [line](https://github.com/cloudquery/cq-provider-aws/blob/main/client/client.go#L382) of detecting which regions are enabled in a specific account is overriding all the accounts. Moreover we don't need the `regions` field on the [client](https://github.com/cloudquery/cq-provider-aws/blob/main/client/client.go#L192) level as this is incorrect and it should be only specified on the account level (even though we can have in the config).

Also, as noted in this [comment](https://github.com/cloudquery/cq-provider-aws/pull/391/files#r779065690) by @bbernays we don't need the list of regions at all, this also remove the burden of maintaining it.

I'll test it manually but also I think we might need some additional testing here as it's a bit hard to verify manually all the edge-cases.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -71,31 +71,6 @@ import (
 
 // Provider Client passed as meta to all table fetchers
 
-var allRegions = []string{
-	"us-east-1",
-	"us-east-2",
-	"us-west-1",
-	"us-west-2",
-	"af-south-1",
-	"ap-east-1",
-	"ap-south-1",
-	"ap-northeast-1",
-	"ap-northeast-2",
-	"ap-southeast-1",
-	"ap-southeast-2",
-	"ca-central-1",
-	"cn-north-1",
-	"cn-northwest-1",
-	"eu-central-1",
-	"eu-west-1",
-	"eu-west-2",
-	"eu-west-3",
-	"eu-south-1",
-	"eu-north-1",
-	"me-south-1",
-	"sa-east-1",
-}
-
 var envVarsToCheck = []string{
 	"AWS_PROFILE",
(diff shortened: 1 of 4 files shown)


yevgenypats:
Closing - Fixed in https://github.com/cloudquery/cq-provider-aws/pull/391

PR #396, opened 2022-01-09
feat: AWS Provider Improvements And Docs

bbernays:
- Improves Docs
- Adds support for each account to grab credentials from a separate local profile
- Removes overloaded `account_id` parameter 
- Adds unit tests

- [ ] Relates OR Closes #0000

```

```


client/client.go
```diff
@@ -324,24 +324,13 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 		// This is a try to solve https://aws.amazon.com/premiumsupport/knowledge-center/iam-validate-access-credentials/
 		// with this https://github.com/aws/aws-sdk-go-v2/issues/515#issuecomment-607387352
 		switch {
```
yevgenypats, line 326: maybe remove this switch and just use standard `if` and `else`

client/client.go
```diff
@@ -361,6 +350,17 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 			return nil, fmt.Errorf(awsFailedToConfigureErrMsg, accountID, err, checkEnvVariables())
 		}
 
+		if account.RoleARN != "" {
```
yevgenypats, line 353: maybe put this in the `switch` or new `if` `else` ?

client/client.go
```diff
@@ -324,24 +324,13 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 		// This is a try to solve https://aws.amazon.com/premiumsupport/knowledge-center/iam-validate-access-credentials/
 		// with this https://github.com/aws/aws-sdk-go-v2/issues/515#issuecomment-607387352
 		switch {
-		case accountID != "default" && account.RoleARN != "":
-			// assume role if specified (SDK takes it from default or env var: AWS_PROFILE)
+		case account.LocalProfile != "":
 			awsCfg, err = config.LoadDefaultConfig(
 				ctx,
 				config.WithDefaultRegion(defaultRegion),
+				config.WithSharedConfigProfile(account.LocalProfile),
```
yevgenypats, line 331: will it still continue to work with `AWS_PROFILE` ?

yevgenypats (changes requested):
Looks great. few nits and I would add integration tests to this part of the code as we don't have such and it's hard to know if we broke something or now without them. 

Also those kind of tests can serve as a great reference for users with all options of configuration for multi-account auth.

client/client.go
bbernays: I was using a switch whenever there was more than 3 or more cases and a if/else when there were fewer than 3 cases.

client/client.go
bbernays: Yes, because it only overrides when `account.LocalProfile` is not an empty string

.github/workflows/cq_fetch.yml
```diff
@@ -161,6 +161,92 @@ jobs:
...
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.ASSUME_ROLE_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.ASSUME_ROLE_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+
+      - name: Fetch
+        run: |
+          rm -rf .cq_reattach
+          echo "[only_us_east_2]\nregion=us-east-2\noutput=json\n[only_us_west_2]\nregion=us-west-2\noutput=json\n" > config
```
yevgenypats, line 239: can we do that as a standard integration test ? via `TestResource` ?

yevgenypats (changes requested):
Looks good Im wandering if it's possible to use `TestResource` instead of the shell/ci test?

.github/workflows/cq_fetch.yml
bbernays: We might be able to move some functionality to the `TestResource` but in order to do this test you have to setup the surrounding environment variables and files

yevgenypats: maybe we can put this under additional test tag? in any case we are providing those variables here.

client/client.go
bbernays: Not sure this comment is still relevant after the refactoring...

client/client.go
```diff
@@ -286,98 +286,106 @@ func isAllRegions(regions []string) bool {
...
+
+	if awsConfig.AWSDebug {
+		awsCfg.ClientLogMode = aws.LogRequest | aws.LogResponse | aws.LogRetries
+		awsCfg.Logger = AwsLogger{logger.With("accountName", account.accountName)}
+	}
+	_, err = awsCfg.Credentials.Retrieve(ctx)
+	if err != nil {
+		return awsCfg, fmt.Errorf(awsFailedToConfigureErrMsg, account.accountName, err, checkEnvVariables())
+	}
+
```
roneli, line 343:
```suggestion
	if 	_, err := awsCfg.Credentials.Retrieve(ctx); err != nil {
		return awsCfg, fmt.Errorf(awsFailedToConfigureErrMsg, account.accountName, err, checkEnvVariables())
	}

```

client/client.go
```diff
@@ -286,98 +286,106 @@ func isAllRegions(regions []string) bool {
...
+		}
+		provider := stscreds.NewAssumeRoleProvider(stsClient, account.RoleARN, opts...)
+		// Test out retrieving credentials
+
+		awsCfg.Credentials = aws.NewCredentialsCache(provider)
+	}
+
+	if awsConfig.AWSDebug {
+		awsCfg.ClientLogMode = aws.LogRequest | aws.LogResponse | aws.LogRetries
+		awsCfg.Logger = AwsLogger{logger.With("accountName", account.accountName)}
```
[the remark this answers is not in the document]
bbernays: We don't have the account id at this time we only get the account id after we setup the aws client and call `get-caller-identity` 


Part of this pr is to remove the accountid field because it was overloaded in meaning.

client/client.go
```diff
@@ -286,98 +286,106 @@ func isAllRegions(regions []string) bool {
...
+		if account.ExternalID != "" {
+			opts = append(opts, func(opts *stscreds.AssumeRoleOptions) {
+				opts.ExternalID = &account.ExternalID
+			})
+		}
+		if stsClient == nil {
+			stsClient = sts.NewFromConfig(awsCfg)
+		}
+		provider := stscreds.NewAssumeRoleProvider(stsClient, account.RoleARN, opts...)
+		// Test out retrieving credentials
```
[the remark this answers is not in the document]
bbernays: yeah a remnant of refactoring for testability

client/client.go
```diff
@@ -286,98 +286,106 @@ func isAllRegions(regions []string) bool {
...
 
 	for _, account := range awsConfig.Accounts {
-		var err error
-		var awsCfg aws.Config
+		logger.Debug("accounts", "account", account)
+		if account.AccountID != "" {
+			return nil, fmt.Errorf("account_id is no longer supported. To specify a profile use `local_profile`. To specify an account alias use `account_name`")
+		}
+		// if accountName is not specified, use block label
+		account.accountName = account.AccountName
```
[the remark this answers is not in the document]
bbernays: Removed `accountName`

client/config.go
```diff
@@ -1,11 +1,14 @@
...
-	ExternalID string   `hcl:"external_id,optional"`
-	Regions    []string `hcl:"regions,optional"`
+	ID           string   `hcl:",label"`
+	AccountID    string   `hcl:"account_id,optional"`
+	AccountName  string   `hcl:"account_name,optional"`
+	RoleARN      string   `hcl:"role_arn,optional"`
+	LocalProfile string   `hcl:"local_profile,optional"`
+	ExternalID   string   `hcl:"external_id,optional"`
+	Regions      []string `hcl:"regions,optional"`
+	accountName  string
```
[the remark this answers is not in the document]
bbernays: I have removed the small accountName

client/client.go
bbernays: fixed

client/client.go
```diff
@@ -286,98 +286,106 @@ func isAllRegions(regions []string) bool {
...
 
 		if len(localRegions) == 0 {
 			localRegions = awsConfig.Regions
 		}
 
-		err = isValidRegions(localRegions)
+		err := isValidRegions(localRegions)
 		if err != nil {
 			return nil, err
 		}
```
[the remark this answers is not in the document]
bbernays: changed from `isValidRegions` to  `verifyRegions`

roneli (approved):
LGTM

PR #401, opened 2022-01-10
Testable Migrations

disq:
Using the new version `4` proto and SDK.
Old migrations are no more. 🦜 
Version is assumed to be `v0.10.0`, migrations version `20`. Migration has been generated with:
```
go run tools/migrations/main.go -prefix 20_v0.10.0. -full 
```

To create a potential "upgrade" migration one needs to run:
```
go run tools/migrations/main.go -dsn 'THE-DSN'
```
and check the generated `unreleased_*.sql` files accordingly.

Please see relevant PRs on other projects:
https://github.com/cloudquery/cq-provider-sdk/pull/146


resources/provider/migrations/postgres/20_v0.11.0.down.sql
```diff
@@ -0,0 +1,562 @@
+-- Autogenerated by migration tool on 2022-01-17 13:44:59
```
roneli, line 1: why is this 20_?

roneli (approved):
LGTM, would add a migration tester in SDK for provider to use to test its migrations

resources/provider/migrations/postgres/20_v0.11.0.down.sql
disq: Temporary future version... we can adjust it to be whatever the next int will be.
Keeping it a round even number might be a good indication for breaking backwards compatibility too

disq: Fixed as 15_

PR #406, opened 2022-01-13
fix: Config Status Recorder

### LINKED ISSUE #407: recording_status of config.configuration_recorders does not reflect actual Config recording status
### Describe the bug

The `recording_status` of the `aws_config_configuration_recorders` table (for `config.configuration_recorders`) does not reflect actual Config recording status.

I have fetched `config.configuration_recorders` data for an AWS account that has Config recording enabled (and recording) in all active AWS regions, but `recording_status` consistently shows up as `false`.

### Expected Behavior

The `recording_status` of the `aws_config_configuration_recorders` table (for `config.configuration_recorders`) should reflect reality at time of data `fetch`.

### Steps to Reproduce

I assume this can be reproduced with any AWS account having Config recording in at least one region.

### Possible Solution

I've also noticed that the `status_last_*` fields in `aws_config_configuration_recorders` are also empty in regions where Config is actually running and recording config.

CQ version: 0.19.0, cq-provider-aws version: 0.9.0

### Additional Context

I have not tested to determine what `aws_config_configuration_recorders.recording_status` shows for regions where Config actually is NOT recording.

bbernays:
- [x] Closes #407 

```

```


resources/services/config/configuration_recorders.go
```diff
@@ -136,11 +137,12 @@ func fetchConfigConfigurationRecorders(ctx context.Context, meta schema.ClientMe
 	for _, configurationRecorder := range resp.ConfigurationRecorders {
 		var configurationRecorderStatus types.ConfigurationRecorderStatus
 		for _, s := range status.ConfigurationRecordersStatus {
-			if s.Name == configurationRecorder.Name {
+			if *s.Name == *configurationRecorder.Name {
```
yevgenypats, line 140: I think we should add a `nil` check

yevgenypats (changes requested):
one comment.

roneli (approved):
LGTM

PR #412, opened 2022-01-18
chore: Linting tags

bbernays:
- [ ] Relates OR Closes #0000

```

```

### DIFF
--- a/resources/services/accessanalyzer/analyzers_mock_test.go
+++ b/resources/services/accessanalyzer/analyzers_mock_test.go
@@ -1,3 +1,4 @@
+//go:build mock
 // +build mock
 
 package accessanalyzer
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,6 +11,9 @@ run:
     - all
 
 linters-settings:
+  gofmt:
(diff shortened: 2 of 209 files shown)


resources/services/accessanalyzer/analyzers_mock_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build mock
 // +build mock
```
amanenk, line 2: Should we remove old tags?

amanenk (approved):
LGTM
Should we use both types of tags or leave only latest variant?

resources/services/accessanalyzer/analyzers_mock_test.go
bbernays: I ran this gofmt command from the root of the repo and this is what it did  `gofmt -s -w .` ... gofmt behavior is usually taken as canonical

bbernays: I have opened an issue to add gofmt to our ci https://github.com/cloudquery/.github/issues/1

disq (approved):

PR #415, opened 2022-01-18
fix: Sagemaker training jobs panic resolved

amanenk:


disq (approved):
Could you also fix the `r.TensorBoardOutputConfig` thing (in `resolveSagemakerTrainingJobTensorBoardOutputConfig`) and `expected <type> but got %T` error messages?

PR #418, opened 2022-01-19
fix: Continue remove account_id in config

bbernays:
- [ ] Relates OR Closes #0000

```

```

### DIFF
--- a/client/config.go
+++ b/client/config.go
@@ -21,11 +21,11 @@ type Config struct {
 func (c Config) Example() string {
 	return ` configuration {
   // Optional. if you want to assume role to multiple account and fetch data from them
-  // accounts "<YOUR ACCOUNT ID>" {
+  // accounts "<UNIQUE ACCOUNT IDENTIFIER>" {
     // Optional. Role ARN we want to assume when accessing this account
     // role_arn = < YOUR_ROLE_ARN >
-    // Optional. Account ID we want to assume when accessing this account - override the block label
-    // account_id = < YOUR ACCOUNT ID >
+    // Optional. Named profile in config or credential file from where CQ should grab credentials
+    // LocalProfile = < PROFILE_NAME >
   // }
   // Optional. by default assumes all regions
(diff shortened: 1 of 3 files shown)


client/config.go
```diff
@@ -21,11 +21,11 @@ type Config struct {
 func (c Config) Example() string {
 	return ` configuration {
   // Optional. if you want to assume role to multiple account and fetch data from them
-  // accounts "<YOUR ACCOUNT ID>" {
+  // accounts "<UNIQUE ACCOUNT IDENTIFIER>" {
     // Optional. Role ARN we want to assume when accessing this account
     // role_arn = < YOUR_ROLE_ARN >
-    // Optional. Account ID we want to assume when accessing this account - override the block label
-    // account_id = < YOUR ACCOUNT ID >
```
roneli, line 28: why? its still a valid attribute in the hcl, I am confused from the previous PR #396

client/testdata/assume_role.hcl
```diff
@@ -17,7 +17,6 @@ provider "aws" {
 
     accounts "default" {
       role_arn = "${ASSUME_ROLE_ARN}"
-      account_id = "${ASSUME_ROLE_ACCOUNT_ID}"
```
roneli, line 20: mistake?

roneli (changes requested):

client/config.go
bbernays: The mistake is that it is still a valid attribute in hcl. fixed now

bbernays: We had already made it an error: https://github.com/cloudquery/cq-provider-aws/blob/2fc0a4d434fb274352cb1c74562b22fc8adb8713/client/client.go#L361-L363

client/testdata/assume_role.hcl
bbernays: No

roneli (approved):

PR #425, opened 2022-01-24
fix: Hotfix for efs filesystems

amanenk:
- added ignore error for backup policies that was not found


roneli (approved):
LGTM

roneli (commented):
nit: missing description for EFS column

PR #427, opened 2022-01-24
fix: Sagemaker panics

disq:
It was failing in [resolveSagemakerTrainingJobDebugHookConfigsCollectionConfigurations](https://github.com/cloudquery/cq-provider-aws/blob/c6d832b/resources/services/sagemaker/training_jobs.go#L640) but I went overboard with the nil checks.

### DIFF
--- a/resources/services/sagemaker/training_jobs.go
+++ b/resources/services/sagemaker/training_jobs.go
@@ -601,7 +601,7 @@ func fetchSagemakerTrainingJobAlgorithmSpecifications(_ context.Context, _ schem
 	r, ok := parent.Item.(*sagemaker.DescribeTrainingJobOutput)
 	if !ok {
 		return fmt.Errorf("expected DescribeTrainingJobOutput but got %T", parent.Item)
-	} else if r == nil || r.AlgorithmSpecification == nil {
+	} else if r.AlgorithmSpecification == nil {
 		return nil
 	}
 	res <- r.AlgorithmSpecification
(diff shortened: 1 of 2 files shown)


resources/services/sagemaker/training_jobs.go
```diff
@@ -751,8 +765,7 @@ func resolveSagemakerTrainingJobProfilerConfig(_ context.Context, _ schema.Clien
 	r, ok := resource.Item.(*sagemaker.DescribeTrainingJobOutput)
 	if !ok {
 		return fmt.Errorf("expected DescribeTrainingJobOutput but got %T", resource.Item)
-	}
-	if r.ProfilerConfig == nil {
+	} else if r == nil || r.ProfilerConfig == nil {
```
roneli, line 768: nit: I suggest this instead of the else if for all the changes here
```suggestion
	if r != nil && r.ProfilerConfig == nil {
```

resources/services/sagemaker/training_jobs.go
```diff
@@ -848,10 +859,12 @@ func resolveSagemakerTrainingJobTags(ctx context.Context, meta schema.ClientMeta
 }
 func resolveSagemakerTrainingJobSecondaryStatusTransitions(_ context.Context, _ schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
 	r, ok := resource.Item.(*sagemaker.DescribeTrainingJobOutput)
-
 	if !ok {
 		return fmt.Errorf("expected DescribeEndpointConfigOutput but got %T", resource.Item)
+	} else if r == nil {
+		return nil
 	}
+
```
roneli, line 867:
```suggestion
	if r == nil {
		return nil
	}

```

roneli (approved):
LGTM, some nitpicks

resources/services/sagemaker/training_jobs.go
disq: This wouldn't cover all cases, e.g. `r` being `nil`.

resources/services/sagemaker/training_jobs.go
disq: if..elseif chains with every "upper" block always returning isn't that confusing, is it? otherwise linter would've caught that.

roneli: Its a a nitpick, I find it harder to look at, and lets less letters, but ill accept also else-if, either way it should be consistent across the whole repo.

roneli (approved):
LGTM

PR #431, opened 2022-01-25
fix: RDS Certificates PK issue and column names

disq:
PK should be the `arn` (or identifier) and not thumbprint. ~Fixed `_till` columns as well.~


resources/services/rds/certificates.go
```diff
@@ -17,7 +17,7 @@ func RdsCertificates() *schema.Table {
 		Multiplex:    client.ServiceAccountRegionMultiplexer("rds"),
 		IgnoreError:  client.IgnoreAccessDeniedServiceDisabled,
 		DeleteFilter: client.DeleteAccountRegionFilter,
-		Options:      schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "thumbprint"}},
+		Options:      schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "arn"}},
```
bbernays, line 20: The thing I had been trying to understand was if thumbprint could be replicated accross region, but this seems to have fixed that issue as region is part of arn

roneli: wouldn't account_id be included in arn? we can just use that instead?

roneli (approved):
LGTM had one extra comment on @bbernays comment

resources/services/rds/certificates.go
disq: correct, fixed

bbernays: So maybe the better PK choice is `arn` and `thumbprint`?

bbernays: Actually nevermind, remembered why we need `account_id`

PR #432, opened 2022-01-25
feat: Remove PII

disq:
I don't want to log/show pii-removed error messages to the end user (unless requested perhaps?) because that would be hard for debugging permissions and general setup. So maybe SDK `diag.Diagnostic` needs a new entry, a `redactedError`? Or just report multiple diag.Diagnostics, one normal and one redacted, and let the SDK choose which one to use when?

### DIFF
--- a/client/errors.go
+++ b/client/errors.go
@@ -14,6 +14,14 @@ import (
 
 func ErrorClassifier(meta schema.ClientMeta, resourceName string, err error) diag.Diagnostics {
 	client := meta.(*Client)
+
+	// Don't override if already a diagnostic, just redact
+	if d, ok := err.(diag.Diagnostic); ok {
+		return diag.Diagnostics{
+			RedactError(client.Accounts, d),
+		}
+	}
+
 	var ae smithy.APIError
 	if errors.As(err, &ae) {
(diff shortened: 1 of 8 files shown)


client/errors.go
```diff
@@ -86,3 +87,19 @@ func isCodeThrottle(code string) bool {
 	_, ok := throttleCodes[code]
 	return ok
 }
+
+var (
+	requestIdRegex = regexp.MustCompile(`RequestID: [0-9a-f-]+`)
+	arnIdRegex     = regexp.MustCompile(`\sarn:aws:.+?\s`)
```
bbernays, line 93: This will only find arn's in the [commercial regions](https://miparnisariblog.wordpress.com/2020/10/26/aws-partitions/). This will not find any arns in: 

`aws-cn`
`aws-us-gov`
`aws-iso`
`aws-iso-b`

disq: Good catch! will fix (`aws.*?:` maybe)

bbernays: Yeah I think that would work

roneli (approved):
LGTM

PR #443, opened 2022-01-26
fix: Added nil check to avoid panics in elvb2 linters and lambda functions

amanenk:
- added some nil checks to avoid panics


roneli (approved):
LGTM

PR #444, opened 2022-01-26
chore: Add test_policy.yml workflow

yevgenypats:
- [ ] Relates OR Closes #0000

```

```

### DIFF
--- a/.github/workflows/test_policy.yml
+++ b/.github/workflows/test_policy.yml
@@ -1,0 +1,68 @@
+name: Test Policy
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+jobs:
+  test_policy:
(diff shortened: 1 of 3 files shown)


.github/workflows/test_policy.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Get dependencies
+        run: |
+          go get -v -t -d ./...
+
+      - name: Build
+        run: go build -v .
```
disq, line 48: Why do we build it if we're going to `go run main.go`? If we're testing buildability, `go vet` would suffice

yevgenypats: you are right. was a copy paste from another action.

PR #446, opened 2022-01-27
fix: Make RDS certificates 'Global'

disq:
I think this is the ticket to fixing the rds certificate errors. [AWS doc](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_Certificate.html)

The `certificate_identifier` is "rds-ca-2019" (which is in the `arn` along with region) and the `type` always seems to be `CA`... And there doesn't seem to be a way of uploading/creating a custom certificate in the UI.


roneli (approved):
LGTM

bbernays:
@disq @roneli - Modify RDS Certs there is a CLI/API:
 https://docs.aws.amazon.com/cli/latest/reference/rds/modify-certificates.html

disq:
Then what we might need is a "global" type for some of the IDs? @roneli

PR #448, opened 2022-01-28
feat: Added generic api error ignore routine

### LINKED ISSUE #445: Add a generic handler of not found errors
### Describe the bug

Different resources has a different `NotFound` status. not found is not an error that should be returned.

### Expected Behavior

don't return errors if resource is not found

### Steps to Reproduce

Run describe ecs cluster on removed cluster

### Possible Solution

Create a generic handler that checks if error is related to not found.

0.19.2; 0.9.2

### Additional Context

_No response_

amanenk:
Most of the errors that should be ignored has a few common prefixes/suffixes. The function checks if error should be ignored and if it is it logs error as warning and allows to skip it. The function is used only in a few places. After PR approve I will propagate it to other resources.
closes #445

### DIFF
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -10,8 +10,9 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/aws/aws-sdk-go-v2/aws/arn"
 	"github.com/aws/smithy-go"
+
+	"github.com/aws/aws-sdk-go-v2/aws/arn"
 	"github.com/cloudquery/cq-provider-sdk/provider/schema"
 )
 
@@ -217,4 +218,29 @@ func ResolveARN(service AWSService, resourceID func(resource *schema.Resource) (
 // Region  and account id are left empty.
 func ResolveARNGlobal(service AWSService, resourceID func(resource *schema.Resource) ([]string, error)) schema.ColumnResolver {
 	return resolveARN(service, resourceID, false, false)
+}
+
+func stringOneOf(a string, args ...interface{}) bool {
+	for _, b := range args {
+		if b == a {
+			return true
+		}
+	}
+	return false
+}
+
+// IgnoreApiError checks if api error should be ignored
+func (c *Client) IgnoreApiError(err error) bool {
+	var ae smithy.APIError
+	if errors.As(err, &ae) {
+		errorCode := ae.ErrorCode()
+		if stringOneOf(errorCode, "ResourceNotFoundException", "WAFNonexistentItemException") ||
+			strings.HasPrefix(errorCode, "NoSuch") ||
+			strings.HasSuffix(errorCode, "NotFound") ||
+			strings.HasSuffix(errorCode, "NotFoundError") {
+			c.logger.Warn("API returned 404 error ignoring it...", "error", err)
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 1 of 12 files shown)


disq:
also: https://github.com/cloudquery/cq-provider-aws/pull/462

client/helpers.go
```diff
@@ -218,3 +219,28 @@ func ResolveARN(service AWSService, resourceID func(resource *schema.Resource) (
...
 }
+
+func stringOneOf(a string, args ...interface{}) bool {
+	for _, b := range args {
+		if b == a {
+			return true
+		}
+	}
+	return false
+}
```
roneli, line 230: use the funk library instead you can use funk.Contains

client/helpers.go
```diff
@@ -218,3 +219,28 @@ func ResolveARN(service AWSService, resourceID func(resource *schema.Resource) (
...
+
+// IgnoreApiError checks if api error should be ignored
+func (c *Client) IgnoreApiError(err error) bool {
```
roneli, line 233: the naming is wrong, its a IsNotFoundError

client/helpers.go
```diff
@@ -218,3 +219,28 @@ func ResolveARN(service AWSService, resourceID func(resource *schema.Resource) (
...
+	var ae smithy.APIError
+	if errors.As(err, &ae) {
+		errorCode := ae.ErrorCode()
+		if stringOneOf(errorCode, "ResourceNotFoundException", "WAFNonexistentItemException") ||
+			strings.HasPrefix(errorCode, "NoSuch") ||
+			strings.HasSuffix(errorCode, "NotFound") ||
+			strings.HasSuffix(errorCode, "NotFoundError") {
```
roneli, line 240: we could use contains here no? or a regex?

client/helpers.go
```diff
@@ -218,3 +219,28 @@ func ResolveARN(service AWSService, resourceID func(resource *schema.Resource) (
...
+	}
+	return false
+}
+
+// IgnoreApiError checks if api error should be ignored
+func (c *Client) IgnoreApiError(err error) bool {
+	var ae smithy.APIError
+	if errors.As(err, &ae) {
+		errorCode := ae.ErrorCode()
+		if stringOneOf(errorCode, "ResourceNotFoundException", "WAFNonexistentItemException") ||
```
roneli, line 237: would make this error codes a const of strings as it might grow larger overtime.

roneli (changes requested):
Good helper method, had some comments

disq (approved):

client/helpers.go
```diff
@@ -218,3 +218,26 @@ func ResolveARN(service AWSService, resourceID func(resource *schema.Resource) (
...
+	if errors.As(err, &ae) {
+		errorCode := ae.ErrorCode()
+		for _, s := range notFoundErrorPrefixes {
+			if strings.Contains(errorCode, s) {
+				c.logger.Warn("API returned \"NotFound\" error ignoring it...", "error", err)
+				return true
+			}
+		}
+	}
+	return false
```
roneli, line 242: Nit:
```suggestion
	if !errors.As(err, &ae) {
	   return false
	}
        errorCode := ae.ErrorCode()
        for _, s := range notFoundErrorPrefixes {
	        if strings.Contains(errorCode, s) {
		        c.logger.Warn("API returned \"NotFound\" error ignoring it...", "error", err)
		        return true
	        }
        }
        return false
```

roneli (approved):
LGTM, had one nit.

roneli (approved):
LGTM

PR #450, opened 2022-01-28
test: Refactor integration tests

yevgenypats:
All integration tests now run from TestIntegration.
This makes the tests faster and remove huge amount of connection
to the database.

Also remove bash magic to test for warnings using grep and do
all the checks on the SDK level.


roneli (approved):
LGTM

PR #482, opened 2022-02-16
feat: Add 'OwnershipControls' Column to S3 resource

shimonp21:


resources/services/s3/buckets.go
```diff
@@ -787,3 +799,41 @@ func resolveBucketTagging(ctx context.Context, meta schema.ClientMeta, resource
...
+
+	ownershipControlRules := getBucketOwnershipControlOutput.OwnershipControls.Rules
+
+	if len(ownershipControlRules) == 0 {
+		return nil
+	}
+
+	if len(ownershipControlRules) > 1 {
+		return errors.New("more than one ownership control rules")
+	}
```
bbernays, line 836: Why can we not support more than 1 rule?

resources/services/s3/buckets.go
```diff
@@ -787,3 +799,41 @@ func resolveBucketTagging(ctx context.Context, meta schema.ClientMeta, resource
...
 	return resource.Set("tags", tags)
 }
+
+func resolveBucketOwnershipControls(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, bucketName, bucketRegion string) error {
+	c := meta.(*client.Client)
+	svc := c.Services().S3
+
+	getBucketOwnershipControlOutput, err := svc.GetBucketOwnershipControls(ctx, &s3.GetBucketOwnershipControlsInput{Bucket: aws.String(bucketName)}, func(options *s3.Options) {
+		options.Region = bucketRegion
+	})
```
bbernays, line 809: Because the list S3 bucket for every region returns the same thing, does this mean we will grab ALL s3 bucket details for every region that the user specifies in the `config.hcl`?

resources/services/s3/buckets.go
shimonp21: Agreed. Let's change it to a text-array

resources/services/s3/buckets.go
bbernays: Not actually an issue

bbernays (approved):
LGTM

PR #484, opened 2022-02-17
chore: Delete pull-request template

shimonp21:
As discussed with Yevgeny, our integration-tests now run in CI - so no need for devs to run them manually on their workstation.
In addition, it is quite unreasonable to expect community devs to have access to an AWS environment with hundreds of different resources,
which is what would be required currently for the integration-tests to pass for them.


yevgenypats:
Approved. This will get an updated MD with link to the contribution once available (should be ready this week)

PR #496, opened 2022-02-23
fix: EC2 Security Group Nested PK Issue

bbernays:
According to the docs about `UserIdGroupPair.UserId`:

```
	// The ID of an Amazon Web Services account. For a referenced security group in
	// another VPC, the account ID of the referenced security group is returned in the
	// response. If the referenced security group is deleted, this value is not
	// returned.
```

So with this PR I am ensuring the value is never `null`

### DIFF
--- a/resources/services/ec2/security_groups.go
+++ b/resources/services/ec2/security_groups.go
@@ -3,6 +3,7 @@ package ec2
 import (
 	"context"
 	"fmt"
+	"math/rand"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ec2"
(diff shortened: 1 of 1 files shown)


disq:
If it's a PK, having more than one deleted user in the same group, with the `DELETED` id might cause problems?

disq:
> If it's a PK, having more than one deleted user in the same group, with the `DELETED` id might cause problems?

erm, maybe ignore the row if user is `nil`? Resulting table should reflect the latest state of the infra anyway.

bbernays:
> erm, maybe ignore the row if user is `nil`? Resulting table should reflect the latest state of the infra anyway.

I am hesitant to drop data. I could imagine users being interested in finding instances like this where they want to find. I see we have two options:
1. Add a random string to the DELETED value for example `DELETED-as2d1`
2. Remove the PK and allow duplicate rows

disq:
Still not sure about this. If it's deleted it shouldn't be in the table. What if it it's deleted for real and there's no more `item` with `nil` userid, is the stale data remover going to capture this and remove the row?

disq:
New approach: Can we make sure the random id part stays persistent as long as it's the same row? Maybe use the user's _index_ in the array and put it as a suffix for `DELETED-` ?

bbernays:
> New approach: Can we make sure the random id part stays persistent as long as it's the same row? Maybe use the user's _index_ in the array and put it as a suffix for `DELETED-` ?

I don't think that we can depend on array ordering. as any new peering connections could alter the ordering

disq:
If the ordering changes they get a new row, and old one is deleted or replaced? Better than just random I think.

PR #499, opened 2022-02-24
feat: Decouple Admin Credentials From Member Account Assumption

### LINKED ISSUE #498: More Flexible Org Account Role Assumption
### Describe the resource.

Right now the org functionality assumes that the same role has access to be able to list all of the accounts is the same role that can assume a role in all of child accounts

### Use Case

If a customer wants to only create trust relationships between specific roles it is quite likely that the 2 actions will require 2 separate roles 

### Additional context

_No response_

bbernays:
Currently we use the credentials in `admin_account` in order to get a list of all accounts in the organization. In the initial implementation we assumed that the member accounts all had trust relations to the same role that we used to grab the accounts.

This pr removes the need for the `admin_account` to have a trust relationship with the member accounts. Users can now override that assumption and use the credentials in the `member_account` in order to assume the specified role in each member account.

Right now I don't love the name of the new attribute `member_account` any recommendations would be appreciated

Update: changed the `member_account` to `member_trusted_principal`


shimonp21:
1. Please also update documentation for the provider.
2. Re: name, I think it's better to be consistent with all our other configs where we call this type of block "account". I think it's confusing to have ~5 places where we use "_account", and one place where it's named "_credentials". When we refactor to credential-terminology, we should replace all usages at once.

roneli (approved):
LGTM, 1 question here you said you changed `member_account` to  `member_trusted_principal` but I didn't see that change in the code. If I missed something, would suggest to have some backwards compatability perhaps or return an error if the old field is used instead of the new one.

bbernays:
@roneli - `member_account` was never a field that has been released. At the beginning of the pr the field in the configuration was called `member_account` now it is called `member_trusted_principal`

https://github.com/cloudquery/cq-provider-aws/blob/5e2aabbbd317d657330e3530f15f5e4f6be941e0/client/config.go#L18

PR #513, opened 2022-03-01
fix: DirectConnect Gateways are global

disq:
A Direct Connect gateway is a globally available resource. You can create the Direct Connect gateway in any Region and access it from all other Regions. The following describe scenarios where you can use a Direct Connect gateway.[[^]](https://docs.aws.amazon.com/directconnect/latest/UserGuide/direct-connect-gateways-intro.html)

### DIFF
--- a/resources/provider/migrations/timescale/20_v0.10.9.up.sql
+++ b/resources/provider/migrations/timescale/20_v0.10.9.up.sql
@@ -1,0 +1,9 @@
+-- Resource: directconnect.gateways
+TRUNCATE aws_directconnect_gateways CASCADE;
+
+ALTER TABLE IF EXISTS aws_directconnect_gateways
+    DROP CONSTRAINT aws_directconnect_gateways_pk;
+ALTER TABLE IF EXISTS aws_directconnect_gateways
+    ADD CONSTRAINT aws_directconnect_gateways_pk PRIMARY KEY (cq_fetch_date,account_id,id);
+
+ALTER TABLE IF EXISTS "aws_directconnect_gateways" DROP COLUMN IF EXISTS "region";
--- a/resources/services/directconnect/gateways.go
+++ b/resources/services/directconnect/gateways.go
@@ -18,7 +18,7 @@ func DirectconnectGateways() *schema.Table {
 		Resolver:      fetchDirectconnectGateways,
 		Multiplex:     client.AccountMultiplex,
 		IgnoreError:   client.IgnoreAccessDeniedServiceDisabled,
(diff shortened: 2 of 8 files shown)


irmatov (approved):

resources/services/directconnect/gateways.go
```diff
@@ -16,10 +16,10 @@ func DirectconnectGateways() *schema.Table {
 		Name:          "aws_directconnect_gateways",
 		Description:   "Information about a Direct Connect gateway, which enables you to connect virtual interfaces and virtual private gateway or transit gateways.",
 		Resolver:      fetchDirectconnectGateways,
-		Multiplex:     client.ServiceAccountRegionMultiplexer("directconnect"),
+		Multiplex:     client.AccountMultiplex,
 		IgnoreError:   client.IgnoreAccessDeniedServiceDisabled,
-		DeleteFilter:  client.DeleteAccountRegionFilter,
-		Options:       schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "region", "id"}},
+		DeleteFilter:  client.DeleteAccountFilter,
+		Options:       schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "id"}},
```
bbernays, line 22: Can we just use ARN instead?

this is the pattern for ARN:
`arn:${Partition}:directconnect::${Account}:dx-gateway/${DirectConnectGatewayId}`

resources/provider/migrations/timescale/20_v0.10.9.up.sql
```diff
@@ -0,0 +1,9 @@
+-- Resource: directconnect.gateways
+TRUNCATE aws_directconnect_gateways CASCADE;
```
bbernays, line 2: Instead can we use a DELETE statement like this:

```
DELETE FROM
    aws_directconnect_gateways a
        USING aws_directconnect_gateways b
WHERE
    a.cq_id < b.cq_id
    AND a.cq_fetch_date = b.cq_fetch_date
    AND a.account_id = b.account_id
    AND a.id = b.id
```

disq: Keeping data this way (when a PK has changed) doesn't work due to the way `cq_id` is calculated (using PKs) and how we remove those rows before inserting.

resources/services/directconnect/gateways.go
disq: We have many resources where using the ARN instead would/might make sense, maybe a bigger discussion.

PR #514, opened 2022-03-01
fix: Fetch each mq_broker_configuration only once, fetch current configuration

disq:


resources/services/mq/brokers.go
```diff
@@ -427,7 +427,24 @@ func fetchMqBrokerConfigurations(ctx context.Context, meta schema.ClientMeta, pa
...
 		return nil
 	}
-	for _, cfg := range broker.Configurations.History {
+
+	list := broker.Configurations.History
+	if broker.Configurations.Current != nil {
+		list = append(list, *broker.Configurations.Current)
+	}
+
+	// History might contain same Id multiple times (maybe with different revisions) but we're only interested in the latest revision of each
```
bbernays, line 436: Why do we only care about the latest one? Isn't it important to show all of the previous configurations?

disq: because `DescribeConfigurationRevision` seems to return raw data... `DescribeConfigurationOutput` has all the fields neatly, but `DescribeConfigurationRevisionOutput` has this instead:

```go
type DescribeConfigurationRevisionOutput struct {

	// Required. The unique ID that Amazon MQ generates for the configuration.
	ConfigurationId *string

	// Required. The date and time of the configuration.
	Created *time.Time

	// Required. The base64-encoded XML configuration.
	Data *string

	// The description of the configuration.
	Description *string

	// Metadata pertaining to the operation's result.
	ResultMetadata middleware.Metadata
}
```

bbernays (approved):
LGTM... We can handle #516 later

PR #522, opened 2022-03-02
chore: Update SDK

disq:
Since summary messages now include the appended error message, we override the embedded error using `WithError` to make it look pretty.


bbernays (approved):
LGTM

PR #523, opened 2022-03-02
fix: Fix ResolveARNWithRegion

bbernays:


disq (approved):

disq:
Should test resources using this resolver and their output before merging.

bbernays:
> Should test resources using this resolver and their output before merging.

Just ran a fetch using the locally built provider and it spot checked the API Gateway resources where this function is actually used and it looks good

PR #536, opened 2022-03-03
fix: Drift: aws_apigateway_usage_plan_keys resource should match the parent resource (usage_plan_id) as well as id

disq:


bbernays (approved):
LGTM

PR #547, opened 2022-03-08
feat: Add CloudFormation stacks

### LINKED ISSUE #494: CloudFormation Stacks
### Describe the resource.

https://docs.aws.amazon.com/sdk-for-go/api/service/cloudformation/

### Use Case

When resources within AWS accounts are created+managed using CloudFormation, the presence or absence of CloudFormation stacks can be key to understanding whether the account is configured as desired.

### Additional context

_No response_

roneli:
Fixes #494

### Acceptance tests:

image


resources/services/cloudformation/stacks.go
```diff
@@ -0,0 +1,335 @@
...
+	for {
+		output, err := svc.DescribeStacks(ctx, &config, func(options *cloudformation.Options) {
+			options.Region = c.Region
+		})
+		if err != nil {
+			return err
+		}
+		if err != nil {
+			return err
+		}
```
bbernays, line 288: Duplication

resources/services/cloudformation/stacks.go
```diff
@@ -0,0 +1,335 @@
...
+				Name:        "region",
+				Description: "The AWS Region of the resource.",
+				Type:        schema.TypeString,
+				Resolver:    client.ResolveAWSRegion,
+			},
+			{
+				Name:     "arn",
+				Type:     schema.TypeString,
+				Resolver: schema.PathResolver("StackId"),
+			},
```
bbernays, line 40: This field seems to be duplicated here and in `id`

bbernays (changes requested):
A few comments. highlighted inline.
What about `[DescribeStackEvents](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudformation#Client.DescribeStackEvents)?

What about Cloudformation StackSets?

yevgenypats (changes requested):
Terraform plan workflow is now available. checkout examples under `terraform/service_name/**` and a bit of docs in https://github.com/cloudquery/cq-provider-aws/blob/main/docs/contributing/development_environment.md#integration-tests (more is coming)

resources/services/cloudformation/stacks.go
```diff
@@ -0,0 +1,335 @@
...
+	return nil
+}
+func resolveStacksTags(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+	r := resource.Item.(types.Stack)
+	tags := map[string]*string{}
+	for _, t := range r.Tags {
+		tags[*t.Key] = t.Value
+	}
+	return resource.Set("tags", tags)
+}
```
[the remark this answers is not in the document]
roneli: different packages different types, only possible with some reflection magic would want to avoid this for now, maybe in go 1.18 with generics.

resources/services/cloudformation/stacks.go
roneli: I added ARN for the sake of having an ARN column, even though the StackId is actually the ARN. maybe in a future version I can remove from stakcId all the arn part and just keep the UUID.

yevgenypats (approved):

roneli:
> A few comments. highlighted inline. What about `[DescribeStackEvents](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudformation#Client.DescribeStackEvents)?
> 
> What about Cloudformation StackSets?

I tried to add minimal amount of resources for starters, we can add it later in a different PR this PR was mostly to showcase cq-gen + the new tf testing.

client/services.go
```diff
@@ -3,6 +3,8 @@ package client
 import (
 	"context"
 
+	"github.com/aws/aws-sdk-go-v2/service/cloudformation"
+
```
disq, line 7: why this empty line?

resources/provider/provider.go
```diff
@@ -3,6 +3,8 @@ package provider
 import (
 	"embed"
 
+	"github.com/cloudquery/cq-provider-aws/resources/services/cloudformation"
+
```
disq, line 7: another empty line

resources/provider/provider.go
```diff
@@ -94,6 +96,7 @@ func Provider() *provider.Provider {
 			"cloudfront.distributions":              cloudfront.CloudfrontDistributions(),
 			"cloudtrail.trails":                     cloudtrail.CloudtrailTrails(),
 			"cloudwatch.alarms":                     cloudwatch.CloudwatchAlarms(),
+			"cloudformation.stacks":                 cloudformation.Stacks(),
```
disq, line 99: ordering is wrong, should be before cloudfront.*, so line 95 presumably

resources/services/cloudformation/stacks.go
```diff
@@ -0,0 +1,342 @@
...
+	c := meta.(*client.Client)
+	svc := c.Services().Cloudformation
+	for {
+		output, err := svc.ListStackResources(ctx, &config, func(options *cloudformation.Options) {
+			options.Region = c.Region
+		})
+		if err != nil {
+			return err
+		}
+		if err != nil {
```
disq, line 331: duplicated err check

resources/services/cloudformation/stacks_mock_test.go
```diff
@@ -0,0 +1,49 @@
+package cloudformation
+
+import (
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/service/cloudformation"
+
```
disq, line 7: another empty line

disq (changes requested):
Some nits. No drift?

resources/provider/migrations/postgres/21_v0.10.12.up.sql
```diff
@@ -1,7 +1,69 @@
...
+CREATE TABLE IF NOT EXISTS "aws_cloudformation_stacks" (
+	"cq_id" uuid NOT NULL,
+	"cq_meta" jsonb,
+	"account_id" text,
+	"region" text,
+	"arn" text,
+	"creation_time" timestamp without time zone,
+	"stack" text,
+	"status" text,
+	"capabilities" text[],
```
bbernays, line 13: This seems to be duplicated in 22_v0.10.12.up.sql

terraform/cloudformation/prod/main.tf
```diff
@@ -0,0 +1,4 @@
+module "demo" {
+  source = "../modules/test"
+  prefix = "cq"
```
bbernays, line 3: The default prefix should be different than the prefix in the local one. If we have it the same then you can only deploy the terraform to a single region (unless you change the prefix)

terraform/cloudformation/modules/test/variables.tf
```diff
@@ -0,0 +1,16 @@
+variable "prefix" {
+  description = "Prefix to use for all name resources"
+  type        = string
+  validation {
+    condition     = length(var.prefix) == 2
+    error_message = "The prefix should be exactly two characters."
+  }
```
bbernays, line 7: I think this will fail when you have `cq-prod`

disq (approved):

PR #546, opened 2022-03-08
fix: Update EC2 EIPs Primary key

zagronitay:

PR #550, opened 2022-03-09
fix: Eip pk fix

amanenk:
- changed primary key of `aws_ec2_eips`

fixes


disq:
https://github.com/cloudquery/cq-provider-aws/pull/546 was here first :(

PR #567, opened 2022-03-15
fix: Handle `GetModelTemplate` errors

zagronitay:


resources/services/apigateway/rest_apis.go
```diff
@@ -924,6 +929,15 @@ func resolveApigatewayRestAPIModelModelTemplate(ctx context.Context, meta schema
 	response, err := svc.GetModelTemplate(ctx, &config, func(options *apigateway.Options) {
 		options.Region = client.Region
 	})
+
+	var ae smithy.APIError
+	if errors.As(err, &ae) {
+		switch ae.ErrorCode() {
+		case "NotFoundException":
```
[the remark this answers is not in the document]
bbernays: I think ontop of `NotFoundException` we should also handle the `BadRequestException`. That way customers that are looking for Models that have issues can just search for 
``` sql
select * from aws_apigateway_rest_api_models where model_template is nil
```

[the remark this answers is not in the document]
disq: @bbernays good idea, but this would also mask any potential issues in the code no?

resources/services/apigateway/rest_apis.go
```diff
@@ -924,6 +927,11 @@ func resolveApigatewayRestAPIModelModelTemplate(ctx context.Context, meta schema
 	response, err := svc.GetModelTemplate(ctx, &config, func(options *apigateway.Options) {
 		options.Region = client.Region
 	})
+
+	if client.IsNotFoundError(err) {
```
disq, line 931: This will panic in `errors.As` inside the fn, move the check inside `if err != nil` :)

disq:
What's the verdict on `BadRequestException` handling?

bbernays:
> What's the verdict on `BadRequestException` handling?

I would be alright settling for just a warning

disq:
> I would be alright settling for just a warning

So returning a diag with `diag.NewBaseError` in case of `BadRequestException` should work @zagronitay

resources/services/apigateway/rest_apis.go
```diff
@@ -914,18 +914,28 @@ func resolveApigatewayRestAPIModelModelTemplate(ctx context.Context, meta schema
...
 	response, err := svc.GetModelTemplate(ctx, &config, func(options *apigateway.Options) {
-		options.Region = client.Region
+		options.Region = cl.Region
 	})
 	if err != nil {
+		if cl.IsNotFoundError(err) {
+			return nil
+		}
+		if cl.IsBadRequestExceptionError(err) {
+			return diag.NewBaseError(err, diag.RESOLVING, diag.WithType(diag.RESOLVING), client.ParseSummaryMessage(err))
```
disq, line 937:
```suggestion
			return diag.NewBaseError(err, diag.RESOLVING, diag.WithSeverity(diag.WARNING), client.ParseSummaryMessage(err))
```

disq (changes requested):

resources/services/apigateway/rest_apis.go
```diff
@@ -914,18 +916,31 @@ func resolveApigatewayRestAPIModelModelTemplate(ctx context.Context, meta schema
...
-		options.Region = client.Region
+		options.Region = cl.Region
 	})
 	if err != nil {
+		if cl.IsNotFoundError(err) {
+			return nil
+		}
+
+		var ae smithy.APIError
+		errorCode := ae.ErrorCode()
```
yevgenypats, line 940: Good that we have integration test :) This code panics as it just accessing nil. 

https://github.com/cloudquery/cq-provider-aws/runs/5598183118?check_suite_focus=true#step:8:2133

yevgenypats (changes requested):

yevgenypats:
Fixed in https://github.com/cloudquery/cq-provider-aws/pull/591

PR #572, opened 2022-03-17
fix: Drop allocation_id from PK of aws_ec2_nat_gateway_addresses

### LINKED ISSUE #571: NAT Gateway: PK Error
### Describe the bug

```
Resource: ec2.nat_gateways Type: Database Severity: Error
    Summary: failed to insert to table "aws_ec2_nat_gateway_addresses": ERROR: null value in column "allocation_id" of relation "aws_ec2_nat_gateway_addresses" violates not-null constraint (SQLSTATE 23502)
    Detail: null value in column "allocation_id" of relation "aws_ec2_nat_gateway_addresses" violates not-null constraint. [Repeated:2]
```

### Expected Behavior

no error

### Steps to Reproduce

run fetch against private NAT Gateway 

`[Public NAT gateway only] The allocation ID of the Elastic IP address that's associated with the NAT gateway.`
https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-nat-gateways.html

### Possible Solution

alter PKs

all?

### Additional Context

_No response_

irmatov:
Closes #571


resources/provider/migrations/postgres/22_v0.12.0.down.sql
```diff
@@ -0,0 +1,7 @@
+-- Autogenerated by migration tool on 2022-03-17 09:22:40
+-- CHANGEME: Verify or edit this file before proceeding
+
+-- Resource: ec2.nat_gateways
+ALTER TABLE aws_ec2_nat_gateway_addresses ALTER COLUMN allocation_id SET NOT NULL;
```
irmatov, line 5: Had to add this manually. Can this be added by the migration tool, @disq

resources/services/ec2/nat_gateways.go
```diff
@@ -132,7 +132,7 @@ func Ec2NatGateways() *schema.Table {
 				Name:        "aws_ec2_nat_gateway_addresses",
 				Description: "Describes the IP addresses and network interface associated with a NAT gateway.",
 				Resolver:    fetchEc2NatGatewayAddresses,
-				Options:     schema.TableCreationOptions{PrimaryKeys: []string{"nat_gateway_cq_id", "allocation_id", "network_interface_id"}},
```
irmatov, line 135: As another option we can drop PK completely on this subtable, but I think it should be fine just without allocation_id. (at least it fetches all our test resources).

resources/provider/migrations/postgres/22_v0.12.0.down.sql
disq: No, migration tool doesn't support Nulls/Uniques.

disq (approved):

disq:
erm, filenames need to be 22_v0.10.13

resources/services/ec2/nat_gateways.go
bbernays: Do we want to add private_ip to the PK?

bbernays (commented):
Should we add a private and public NAT gateway to our terraform? https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/nat_gateway

disq:
> Should we add a private and public NAT gateway to our terraform? [registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/nat_gateway](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/nat_gateway)

It might not be straightforward as there's an optional PK (depending on connectivity_type) I think... besides it's there in the drift.hcl already, uses default values. Maybe a fix might be in order later

PR #586, opened 2022-03-18
Test: Reporduce hang on pgx

yevgenypats:
To run:` go clean -testcache && go test -run=TestLoad -v ./resources/provider/...`

### DIFF
--- a/resources/provider/load_test.go
+++ b/resources/provider/load_test.go
@@ -1,0 +1,126 @@
+package provider
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/cloudquery/cq-provider-aws/client"
+	"github.com/cloudquery/cq-provider-sdk/provider"
+	"github.com/cloudquery/cq-provider-sdk/provider/schema"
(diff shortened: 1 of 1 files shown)


resources/provider/load_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func fetchLoadTestResource(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	generatedResult := make([]LoadTest, 10000)
+	for i := range generatedResult {
+		generatedResult[i].Arn = uuid.New().String()
+		generatedResult[i].CertificateAuthorityArn = uuid.New().String()
+		generatedResult[i].Type = "random type"
+	}
+	res <- generatedResult
```
bbernays, line 124: Maybe also include a random sleep between 100 ms and 2 seconds to simulate network?

disq:
Does it still hang/timeout if you run it with `-timeout=longtimeout`? like:
```
go clean -testcache && go test -run=TestLoad -timeout=30m -v ./resources/provider/... 2>&1
```

PR #588, opened 2022-03-19
fix: Remove unnecessary warn messages

yevgenypats:
Especially around S3 API calls some properties return NotFound
when they are not set instead of empty result. This is completly
"normal" for these AWS APIs and no reason to log those messages.

### DIFF
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -236,6 +236,7 @@ func (c *Client) IsNotFoundError(err error) bool {
 	errorCode := ae.ErrorCode()
 	for _, s := range notFoundErrorPrefixes {
 		if strings.Contains(errorCode, s) {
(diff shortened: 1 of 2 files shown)


client/helpers.go
```diff
@@ -236,9 +236,16 @@ func (c *Client) IsNotFoundError(err error) bool {
 	errorCode := ae.ErrorCode()
 	for _, s := range notFoundErrorPrefixes {
 		if strings.Contains(errorCode, s) {
-			c.logger.Warn("API returned \"NotFound\" error ignoring it...", "error", err)
```
disq, line 239: maybe keep the log line and use IsAWSErr in line 238? since we're converting normal/expected cases to IsAWSErr usage anyway?

yevgenypats: agreed. I missed that one I think.

PR #589, opened 2022-03-19
refactor: Remove type assertion

yevgenypats:
There is no need for type assertion in resolvers for the following reasons:
* They should be caught in unit-tests
* Even if they happen we catch panics and they are more useful with line numbers
* If we do use type assertion checks we will need to add diag.WrapError
* Moreover this actually hide crucial information like what was the type? as this is crucial to debug and understand what happens.

This removes a lot of unnecessary boilerplate-code which also gives bad example
for future contributions.

p.s - this was done with regex in vscode :) 

```
Search:
([a-zA-Z_]*), ok := (.*)\n[ \t]*if !ok \{\n.*\n[ \t]*\}
Replace:
$1 := $2
```

### DIFF
--- a/resources/services/apigateway/domain_names.go
+++ b/resources/services/apigateway/domain_names.go
@@ -2,7 +2,6 @@ package apigateway
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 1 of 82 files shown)


disq (approved):
nice regex. I didn't check all resources/replacements so hoping it works for all

PR #597, opened 2022-03-21
fix: Check regions

### LINKED ISSUE #583: Provider Hangs if Region name is misspelled
### Describe the bug

Provider hangs if region specified in `config.hcl` is not actually a valid region

### Expected Behavior

Should throw an error

### Steps to Reproduce

run fetch with this config:
``` hcl
provider "aws" {
  configuration {
    regions = ["us-eadst-2"]
  }
  resources = ["*"]
}

```

### Possible Solution

_No response_

all?

### Additional Context

_No response_

zagronitay:


client/client.go
```diff
@@ -268,6 +273,9 @@ func verifyRegions(regions []string) error {
 		if i > 0 && hasWildcard {
 			return errInvalidRegion
 		}
+		if ok := availableRegions[region]; !ok {
+			return fmt.Errorf("unknow region: %s", region)
```
disq, line 277:
```suggestion
			return fmt.Errorf("unknown region: %q", region)
```

client/helpers.go
```diff
@@ -91,6 +91,32 @@ func isSupportedServiceForRegion(service string, region string) bool {
...
 
+func getAvailableRegions() (map[string]bool, error) {
+	readOnce.Do(func() {
+		supportedServiceRegion = readSupportedServiceRegions()
+	})
+
+	regionsSet := make(map[string]bool)
+
+	if supportedServiceRegion == nil {
+		return nil, fmt.Errorf("cloud not get AWS regions/services data")
```
disq, line 102:
```suggestion
		return nil, fmt.Errorf("could not get AWS regions/services data")
```

client/helpers.go
```diff
@@ -91,6 +91,32 @@ func isSupportedServiceForRegion(service string, region string) bool {
...
+	}
+
+	if supportedServiceRegion.Partitions == nil {
+		return nil, fmt.Errorf("cloud not found any AWS partitions")
```
disq, line 106:
```suggestion
		return nil, fmt.Errorf("could not find any AWS partition")
```

disq (changes requested):

bbernays:
Could you add a test for this change?

roneli (approved):
LGTM, would add a tests as suggest by @bbernays

disq (approved):

bbernays (approved):
LGTM

PR #596, opened 2022-03-21
fix: Access analyzer migrations fix

amanenk:
- fixed duplication of migration for `aws_access_analyzer_analyzer_archive_rules` in 23 and 24 migrations files
- removed duplicate of 22 and 23 files

PR #599, opened 2022-03-21
fix: Migrations fix

amanenk:


yevgenypats (changes requested):
Looks like there is some conflicts. Looks good otherwise. 

If it was an issue before that is there a way to add a test for that ?

disq:
#596 is already merged, it's not working?

PR #634, opened 2022-03-28
feat: Added AWS workspaces

### LINKED ISSUE #603: Support AWS Workspaces
### Describe the resource.

"Secure, reliable, and scalable access to persistent desktops from any location"

### Use Case

Understand what resources are provisioned and the configs they have

### Additional context

_No response_

zagronitay:


resources/services/workspaces/directories.go
```diff
@@ -0,0 +1,243 @@
...
+func Directories() *schema.Table {
+	return &schema.Table{
+		Name:         "aws_workspaces_directories",
+		Description:  "Describes a directory that is used with Amazon WorkSpaces.",
+		Resolver:     fetchWorkspacesDirectories,
+		Multiplex:    client.ServiceAccountRegionMultiplexer("workspaces"),
+		IgnoreError:  client.IgnoreAccessDeniedServiceDisabled,
+		DeleteFilter: client.DeleteAccountRegionFilter,
+		Options:      schema.TableCreationOptions{PrimaryKeys: []string{"id"}},
+		Columns: []schema.Column{
```
bbernays, line 22: Can you add an ARN field?
https://docs.aws.amazon.com/workspaces/latest/adminguide/workspaces-access-control.html

resources/services/workspaces/workspaces.go
```diff
@@ -0,0 +1,177 @@
...
+func Workspaces() *schema.Table {
+	return &schema.Table{
+		Name:         "aws_workspaces_workspaces",
+		Description:  "Describes a WorkSpace.",
+		Resolver:     fetchWorkspacesWorkspaces,
+		Multiplex:    client.ServiceAccountRegionMultiplexer("workspaces"),
+		IgnoreError:  client.IgnoreAccessDeniedServiceDisabled,
+		DeleteFilter: client.DeleteAccountRegionFilter,
+		Options:      schema.TableCreationOptions{PrimaryKeys: []string{"id"}},
+		Columns: []schema.Column{
```
bbernays, line 24: Can you add an ARN field?
https://docs.aws.amazon.com/workspaces/latest/adminguide/workspaces-access-control.html

bbernays (changes requested):

resources/services/workspaces/workspaces.go
```diff
@@ -175,3 +182,8 @@ func resolveWorkspacesModificationStates(_ context.Context, _ schema.ClientMeta,
 	}
 	return resource.Set(c.Name, data)
 }
+func ResolveWorkspacesWorkspaceArn(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+	cl := meta.(*client.Client)
+	workspace := resource.Item.(types.Workspace)
+	return resource.Set(c.Name, client.GenerateResourceARN("workspaces", "workspace", *workspace.WorkspaceId, cl.Region, cl.AccountID))
+}
```
bbernays, line 189: Isn't this trying to set the `name` attribute to the ARN value?

Could we do something like this: https://github.com/cloudquery/cq-provider-aws/blob/70e67cf0aba63326db8f605caa0a86e8c41aa85b/resources/services/apigateway/api_keys.go#L42-L43

bbernays (approved):

PR #638, opened 2022-03-30
fix: Remove explicit PK

### LINKED ISSUE #497: `UserId` is Blank if SG has been deleted
### Describe the bug

UserId is a PK in the `aws_ec2_security_group_ip_permission_user_id_group_pairs` table.

It is possible for value to be nil/null and then CQ cannot persist to db as it violates a PK constraint

```go
type UserIdGroupPair struct {
	.
	.
	.
	// The ID of an Amazon Web Services account. For a referenced security group in
	// another VPC, the account ID of the referenced security group is returned in the
	// response. If the referenced security group is deleted, this value is not
	// returned. [EC2-Classic] Required when adding or removing rules that reference a
	// security group in another Amazon Web Services account.
	UserId *string
	.
	.
	.

}
```

### Expected Behavior

store in db

### Steps to Reproduce

delete SG after it VPC has been shared

### Possible Solution

_No response_

all

### Additional Context

_No response_

zagronitay:


irmatov (approved):

bbernays (changes requested):
Don't you need the migrations?

bbernays (approved):

disq (approved):

PR #649, opened 2022-04-04
feat: Add Backup plans, vaults

### LINKED ISSUE #618: Add Support for AWS::Backup::*
- [x] AWS::Backup::BackupPlan
- [x] AWS::Backup::BackupSelection
- [x] AWS::Backup::BackupVault

irmatov:
Closes #618


client/services.go
```diff
@@ -598,3 +599,12 @@ type WorkspacesClient interface {
 	DescribeWorkspaces(ctx context.Context, params *workspaces.DescribeWorkspacesInput, optFns ...func(*workspaces.Options)) (*workspaces.DescribeWorkspacesOutput, error)
 	DescribeWorkspaceDirectories(ctx context.Context, params *workspaces.DescribeWorkspaceDirectoriesInput, optFns ...func(*workspaces.Options)) (*workspaces.DescribeWorkspaceDirectoriesOutput, error)
 }
+
+//go:generate mockgen -package=mocks -destination=./mocks/backup.go . BackupClient
+type BackupClient interface {
```
roneli, line 604: Missing ListTags API https://docs.aws.amazon.com/aws-backup/latest/devguide/API_ListTags.html this for vault/plan/recovery point resoruces.

resources/services/backup/vaults.go
```diff
@@ -0,0 +1,117 @@
...
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/backup"
+	"github.com/cloudquery/cq-provider-aws/client"
+	"github.com/cloudquery/cq-provider-sdk/provider/diag"
+	"github.com/cloudquery/cq-provider-sdk/provider/schema"
+)
+
+func Vaults() *schema.Table {
```
roneli, line 13: What about:
https://docs.aws.amazon.com/aws-backup/latest/devguide/API_GetBackupVaultAccessPolicy.html
https://docs.aws.amazon.com/aws-backup/latest/devguide/API_GetBackupVaultNotifications.html
https://docs.aws.amazon.com/aws-backup/latest/devguide/API_ListRecoveryPointsByBackupVault.html

client/services.go
roneli, line 604: I know this isn't part of the original issue but I would add https://docs.aws.amazon.com/aws-backup/latest/devguide/API_ListProtectedResources.html @bbernays thoughts?

roneli, line 604: This is a global table API https://docs.aws.amazon.com/aws-backup/latest/devguide/API_DescribeGlobalSettings.html
https://docs.aws.amazon.com/aws-backup/latest/devguide/API_DescribeRegionSettings.html

resources/services/backup/plans.go
```diff
@@ -0,0 +1,344 @@
...
+
+func Plans() *schema.Table {
+	return &schema.Table{
+		Name:         "aws_backup_plans",
+		Description:  "Contains metadata about a backup plan.",
+		Resolver:     fetchBackupPlans,
+		Multiplex:    client.ServiceAccountRegionMultiplexer("backup"),
+		IgnoreError:  client.IgnoreAccessDeniedServiceDisabled,
+		DeleteFilter: client.DeleteAccountRegionFilter,
+		Options:      schema.TableCreationOptions{PrimaryKeys: []string{"arn"}},
```
zagronitay, line 23: I think we should prefer id over arn (if it's globally unique)

roneli (changes requested):
Looks good overall, had comments on missing APIs, ill let also @bbernays chime his thoughts around the missing calls.

zagronitay (dismissed):

client/services.go
bbernays: I agree that we should add it. It definitely seems like information that would want to be searched

resources/services/backup/plans.go
bbernays: I would leave it as ARN. When they say global it could mean global to that account or that partition, not global to all of AWS

client/services.go
bbernays: For the sake of moving things forward, if you would prefer we can open up an issue to add support for this if we just want to go ahead and move this PR along

irmatov: Added tags.

resources/services/backup/vaults.go
irmatov: Added, thanks!

client/services.go
roneli: Was an issue opened for this? @irmatov ?

roneli: Would open an issue for this as well.

roneli (approved):
LGTM, would maybe change the migration version to v0.10.16 as that is the next version instead of v0.10.17

resources/services/backup/plans.go
```diff
@@ -0,0 +1,383 @@
...
+	return diag.WrapError(resource.Set(c.Name, tags))
+}
+
+func fetchBackupSelections(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+	plan := parent.Item.(backup.GetBackupPlanOutput)
+	cl := meta.(*client.Client)
+	svc := cl.Services().Backup
+	params := backup.ListBackupSelectionsInput{
+		BackupPlanId: plan.BackupPlanId,
+		MaxResults:   aws.Int32(100),
```
bbernays, line 316: Is this the max that the endpoint supports?

irmatov: SDK does not specify any bounds. Some services are fine without the limit (using some default value), some require it explicitly, so I just picked 100 as a "good enough" value.

resources/services/backup/plans.go
```diff
@@ -0,0 +1,383 @@
...
+					{
+						Name:        "conditions",
+						Description: "A list of conditions that you define to assign resources to your backup plans using tags.",
+						Type:        schema.TypeJSON,
+						Resolver:    resolveSelectionConditions,
+					},
+					{
+						Name:        "list_of_tags",
+						Description: "A list of conditions that you define to assign resources to your backup plans using tags.",
+						Type:        schema.TypeJSON,
```
bbernays, line 219: I think this should be an array rather than JSON

resources/services/backup/vaults.go
```diff
@@ -0,0 +1,389 @@
...
+	params := backup.ListTagsInput{ResourceArn: rp.RecoveryPointArn}
+	tags := make(map[string]string)
+	for {
+		result, err := svc.ListTags(ctx, &params, func(o *backup.Options) { o.Region = cl.Region })
+		if result == nil {
+			break
+		}
+		if err != nil {
+			return diag.WrapError(err)
+		}
```
bbernays, line 379: Shouldn't this be switched? So we check the error first?

resources/services/backup/vaults.go
```diff
@@ -0,0 +1,389 @@
...
+	)
+	if err != nil {
+		var ae smithy.APIError
+		if !errors.As(err, &ae) {
+			return diag.WrapError(err)
+		}
+		if ae.ErrorCode() == "ERROR_2106" {
+			// trying to ignore "ERROR_2106: Failed reading notifications from database for Backup vault ..."
+			return nil
+		}
```
bbernays, line 327: What is this?

resources/services/backup/plans.go
bbernays: [The max is 1000](https://docs.aws.amazon.com/aws-backup/latest/devguide/API_ListBackupSelections.html), Can we set it to that? I would rather try and minimize API requests

irmatov: Great, I'll set it to 1000.

resources/services/backup/plans.go
irmatov: I think it should be renamed to just `tags` and be kept as a json (map) to be consistent with all other resources.

resources/services/backup/vaults.go
irmatov: Sure, my bad.

resources/services/backup/vaults.go
irmatov: if there are no notifications configured for a vault this API request returns that error so it can't be matched with `client.IsNotFoundError`

resources/services/backup/plans.go
bbernays: But it is explicitly just a list of strings... I would prefer that we use the most specific datatype so that we don't have to check if it is a singular value, or object. We can allow the DB to do the type checking for us

bbernays: Nevermind, you are right. It should be a json!

bbernays: I think the name as it is is correct

irmatov: Oh, I misunderstood what the issue is here.

I think the field is fine as it is. We cannot make it an array of anything. We could only make it a relation but I am not sure if it is that useful.

resources/services/backup/vaults.go
irmatov: If there is no notifications attached to a vault then we have this weird error from the API which we cannot match with the already existing `client.IsNotFoundError`.

bbernays (approved):

client/services.go
bbernays: https://github.com/cloudquery/cq-provider-aws/issues/674

bbernays: https://github.com/cloudquery/cq-provider-aws/issues/674

PR #671, opened 2022-04-07
fix: PKs for Virtual MFA Device

bbernays:

PR #672, opened 2022-04-08
fix: Cloudformation stacks fixed path resolver

amanenk:
- fixed path resolver for cloudformation stacks


bbernays:
I think this PR can be closed as the fix was merged in [#671](https://github.com/cloudquery/cq-provider-aws/pull/671)

PR #686, opened 2022-04-12
chore: Add WAF Regional TF resources

irmatov:


spangenberg (approved):
LGTM

terraform/wafregional/prod/terraform.tf
```diff
@@ -0,0 +1,7 @@
+terraform {
+  backend "s3" {
+    bucket         = "cq-provider-aws-tf"
+    key            = "backup"
```
bbernays, line 4: I think this should be `wafregional`

bbernays (changes requested):

terraform/wafregional/prod/terraform.tf
irmatov: You're right. :facepalm: Thanks!

bbernays (approved):

PR #690, opened 2022-04-12
fix: Redact "Encoded authorization failure message"

disq:
This would fix the Sentry issue as these errors can get squashed (diag side) and grouped (sentry side) now.


bbernays (approved):
LGTM

PR #693, opened 2022-04-12
feat: Reduce ECS API Calls

bbernays:
This reduces the number of calls by 50% as tags don't need to be grabbed again.

More work can be done to parallelize the `fetchEcsTaskDefinition`

### DIFF
--- a/resources/services/ecs/task_definitions.go
+++ b/resources/services/ecs/task_definitions.go
@@ -594,7 +594,7 @@ type TaskDefinitionWrapper struct {
 	Tags []types.Tag
 }
 
-func fetchEcsTaskDefinitions(ctx context.Context, res chan<- interface{}, svc client.EcsClient, region, taskArn string) error {
+func fetchEcsTaskDefinition(ctx context.Context, res chan<- interface{}, svc client.EcsClient, region, taskArn string) error {
 	describeClusterOutput, err := svc.DescribeTaskDefinition(ctx, &ecs.DescribeTaskDefinitionInput{
 		TaskDefinition: aws.String(taskArn),
 		Include:        []types.TaskDefinitionField{types.TaskDefinitionFieldTags},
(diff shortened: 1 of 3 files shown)


resources/services/ecs/task_definitions.go
```diff
@@ -588,12 +589,38 @@ func EcsTaskDefinitions() *schema.Table {
 	}
 }
 
+type TaskDefinitionWrapper struct {
+	*types.TaskDefinition
+	Tags []types.Tag
+}
+
+func fetchEcsTaskDefinition(ctx context.Context, res chan<- interface{}, svc client.EcsClient, region, taskArn string) error {
+	describeClusterOutput, err := svc.DescribeTaskDefinition(ctx, &ecs.DescribeTaskDefinitionInput{
```
roneli, line 598: describeClusterOutput?

resources/services/ecs/task_definitions.go
```diff
@@ -606,17 +633,12 @@ func fetchEcsTaskDefinitions(ctx context.Context, meta schema.ClientMeta, parent
 		if len(listClustersOutput.TaskDefinitionArns) == 0 {
 			return nil
 		}
+
 		for _, t := range listClustersOutput.TaskDefinitionArns {
-			describeClusterOutput, err := svc.DescribeTaskDefinition(ctx, &ecs.DescribeTaskDefinitionInput{TaskDefinition: &t}, func(o *ecs.Options) {
-				o.Region = region
-			})
+			err := fetchEcsTaskDefinition(ctx, res, svc, region, t)
```
roneli, line 638: Would maybe make this call in parallel

roneli (commented):
Great fix, had a few comments

resources/services/ecs/task_definitions.go
bbernays: I agree that would be an improvement, I was trying trying to get this out as quickly as possible as this would reduce fetch times by 50%

resources/services/ecs/task_definitions.go
bbernays: fixed

roneli (approved):
LGTM

PR #701, opened 2022-04-13
feat: Parallelize ECS Task Definition Fetch

bbernays:
I created 70 task definitions. Previous version took 8 seconds now it takes 3 second


disq (approved):
LGTM

roneli (approved):
LGTM, would maybe add a few more task definitions in our e2e test, to check that more than 3-5 tasks it works well and it doesn't get stuck.

PR #696, opened 2022-04-13
feat: Add AutoScaling ScheduledActions

spangenberg:
Fixes #615

PR #703, opened 2022-04-14
chore: Add AutoScaling ScheduledActions TF Resources

spangenberg:
Prereq for #696

### DIFF
--- a/terraform/autoscaling/modules/test/autoscaling.tf
+++ b/terraform/autoscaling/modules/test/autoscaling.tf
@@ -1,0 +1,44 @@
+data "aws_ami" "ubuntu" {
+  most_recent = true
+
+  filter {
+    name   = "name"
+    values = ["ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-*"]
+  }
+
+  filter {
+    name   = "virtualization-type"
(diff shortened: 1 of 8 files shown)


terraform/autoscaling/modules/test/autoscaling.tf
```diff
@@ -0,0 +1,44 @@
...
+    values = ["hvm"]
+  }
+
+  owners = ["099720109477"] # Canonical
+}
+
+resource "aws_launch_configuration" "launch_configuration" {
+  name          = "${var.prefix}-launch-configuration"
+  image_id      = data.aws_ami.ubuntu.id
+  instance_type = "t2.micro"
```
bbernays, line 20: Can we use something newer like t3 or t4?

spangenberg: Sure, but since the desired count is `0` is shouldn't matter at all.

bbernays (approved):

terraform/autoscaling/modules/test/autoscaling.tf
spangenberg: Updated nevertheless, should familiarize me anyhow with the new types out in the world :)

PR #712, opened 2022-04-18
chore: Add SSM association TF

irmatov:


terraform/ssm/modules/test/ssm.tf
```diff
@@ -26,3 +26,15 @@ DOC
     Name = "docs"
   }
 }
+
+resource "aws_ssm_association" "test_associations" {
+  name = "AmazonCloudWatch-ManageAgent"
```
zagronitay, line 31: name should have prefix

terraform/ssm/modules/test/ssm.tf
```diff
@@ -26,3 +26,15 @@ DOC
...
+  compliance_severity = "UNSPECIFIED"
+  max_concurrency = 1
+  max_errors = 10
+  association_name = "${var.prefix}-ssm-assoc"
+  targets {
+    key    = "tag:Environment"
+    values = ["Development"]
+  }
```
zagronitay, line 39: please add tag if acceptable

zagronitay (changes requested):

terraform/ssm/modules/test/ssm.tf
irmatov: this is not a resource name (resource name is below as association_name). This is a SSM document name and should reference existing resource.

terraform/ssm/modules/test/ssm.tf
irmatov: tags are not supported for associations

zagronitay (approved):

PR #719, opened 2022-04-19
fix: Broken migrations

roneli:


disq:
Would change down migrations as well.

zagronitay (approved):
LGTM

Issue #621, opened 2022-03-24
Ensure Support for AWS::ElasticBeanstalk::*


bbernays:
- [ ] AWS::ElasticBeanstalk::Application
- [ ] AWS::ElasticBeanstalk::ApplicationVersion
- [ ] AWS::ElasticBeanstalk::Environment

shimonp21:
Adding an error I am running into related to elasticbeanstalk:
![image](https://user-images.githubusercontent.com/37939765/164225030-c32f0285-0a30-4c01-876d-6078a4588478.png)

steps to reproduce:
- authenticate with cq-playground account.
- `cloudquery init aws`
- `cloudquery fetch`

erezrokah:
Thanks @shimonp21, looks like these resources are terminated, but somehow still available via the API:

```
aws elasticbeanstalk describe-environments --region us-east-1

{
    "Environments": [
        {
            "EnvironmentName": "<name-hidden>",
            "EnvironmentId": "*********",
            ...
        }
    ]
}

aws elasticbeanstalk terminate-environment --region us-east-1 --environment-name "<name-hidden>"

An error occurred (InvalidParameterValue) when calling the TerminateEnvironment operation: No Environment found for EnvironmentName = '<name-hidden>'.
```

When I first ran `aws elasticbeanstalk describe-environments --region us-east-1` it printed 2 environments, now it prints 1 😕  I think there might be a delay from the time the resource it destroyed until the API reflects it.

From https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/using-features.terminating.html:
>It takes a few minutes for Elastic Beanstalk to terminate the AWS resources that are running in the environment.

We should probably handle that error in the code.

PR #746, opened 2022-04-21
fix: Classifier redaction from EBS

disq:
Also improve Request ID regex

EBS AccessDenied error still there.

### DIFF
--- a/client/errors.go
+++ b/client/errors.go
@@ -153,7 +153,7 @@ func isCodeThrottle(code string) bool {
 }
 
 var (
-	requestIdRegex = regexp.MustCompile(`\s(RequestID:|request id:|Request ID:)\s[A-Za-z0-9-]+`)
+	requestIdRegex = regexp.MustCompile(`\s([Rr]equest[ _]{0,1}(ID|Id|id):)\s[A-Za-z0-9-]+`)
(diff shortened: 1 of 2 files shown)


erezrokah (approved):
Related to https://github.com/cloudquery/cq-provider-aws/issues/621. I'll look at EBS AccessDenied error as a part of https://github.com/cloudquery/cq-provider-aws/issues/621.

Thanks @disq for handling this 🎉

client/errors.go
```diff
@@ -153,13 +153,14 @@ func isCodeThrottle(code string) bool {
 }
 
 var (
-	requestIdRegex = regexp.MustCompile(`\s(RequestID:|request id:)\s[A-Za-z0-9-]+`)
+	requestIdRegex = regexp.MustCompile(`\s(RequestID:|request id:|Request ID:)\s[A-Za-z0-9-]+`)
```
roneli, line 156: nit: maybe make the regexes case insensitive?

roneli (approved):
LGTM

client/errors.go
disq: So `ReQuEsT iD` can also pass? It's too much I think 😂

disq: but I fixed the regex for you

roneli (approved):

PR #752, opened 2022-04-21
chore: Update WAFv2 Managed Groups to use multiplexer

spangenberg:


disq:
Other WAFv2 also has this stuff I think

disq (approved):
LGTM

disq:
"chore" or "fix" though

PR #755, opened 2022-04-22
fix: Upgrade SDK to return diagnostics in Configure

disq:
goes with https://github.com/cloudquery/cq-provider-sdk/pull/226

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -464,7 +464,7 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 				}
 			})
 		if err != nil {
-			return nil, diags.Add(diag.FromError(fmt.Errorf("failed to find disabled regions for account %s. AWS Error: %w", account.AccountName, err), diag.ACCESS))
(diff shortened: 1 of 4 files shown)


bbernays (approved):

client/client.go
```diff
@@ -461,17 +464,17 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 				}
 			})
 		if err != nil {
-			return nil, fmt.Errorf("failed to find disabled regions for account %s. AWS Error: %w", account.AccountName, err)
+			return nil, diags.Add(diag.FromError(fmt.Errorf("failed to find disabled regions for account %s. AWS Error: %w", account.AccountName, err), diag.ACCESS))
```
roneli, line 467: would maybe add a Details to all diag.Access errors to help users on how to fix the issue.

disq: Fixed for this one (not sure about the msg), but for `loadOrgAccounts`, since it uses a diverse number of calls, all helper functions inside would need to be refactored to return diags themselves and set the details there.

spangenberg (approved):
LGTM

PR #751, opened 2022-04-21
feat: Add AWS::ElasticBeanstalk::ApplicationVersion

erezrokah:
Fixes https://github.com/cloudquery/cq-provider-aws/issues/621

Ensured existing logic works, and added support for `AWS::ElasticBeanstalk::ApplicationVersion`

## Todo
- [x] Handle the bug reported here https://github.com/cloudquery/cq-provider-aws/issues/621#issuecomment-1103846841

## Other considerations
- [ ] Should we create a relation from applications to application versions?

PR #763, opened 2022-04-24
chore: Update make file, add db migration script

erezrokah:
Extracted from https://github.com/cloudquery/cq-provider-aws/pull/751

Aligns naming conventions in our make file and adds a new target for db-migration

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,14 +7,14 @@ install-cq:
 	chmod a+x cloudquery
 
 # start a timescale db running in a local container
-.PHONY: ts-start
-ts-start:
+.PHONY: start-ts
+start-ts:
 	docker run -p 5433:5432 -e POSTGRES_PASSWORD=pass -d timescale/timescaledb:latest-pg14
 
 # stop the timescale db running in a local container
-.PHONY: ts-stop
-ts-stop:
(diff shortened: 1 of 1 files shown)


erezrokah:
Also, if we think it's a good idea I can port these changes to other relevant providers

Makefile
```diff
@@ -7,14 +7,14 @@ install-cq:
 	chmod a+x cloudquery
 
 # start a timescale db running in a local container
-.PHONY: ts-start
-ts-start:
+.PHONY: start-ts
+start-ts:
 	docker run -p 5433:5432 -e POSTGRES_PASSWORD=pass -d timescale/timescaledb:latest-pg14
```
bbernays, line 12: So I actually found that `<action>-<db-flavor>` is harder to use than `<db-flavor>-<action>` because once you start working with a specific db-flavor you are going to want to continue to use it

erezrokah: Good point! I'll update the PR to use the `<db-flavor>-<action>` format

erezrokah: Done in https://github.com/cloudquery/cq-provider-aws/pull/763/commits/11daa6141b7b973937c461d9b7a2a4295695ece5

bbernays (approved):

PR #764, opened 2022-04-24
chore: Add local terraform variables

### LINKED ISSUE #749: Ensure each `terraform/<service>/local` has a `variables.tf` with a prefix var
Extracted from https://github.com/cloudquery/cq-provider-aws/pull/723#discussion_r852944815

To ensure we can run integration tests locally for each provider, we need to support deploying Terraform using a prefix.
In order to use a prefix we need to configure that variable under the `local` folder for each service.

See example https://github.com/cloudquery/cq-provider-aws/tree/main/terraform/xray/local

erezrokah:
Fixes https://github.com/cloudquery/cq-provider-aws/issues/749


terraform/wafregional/local/.terraform.lock.hcl
```diff
@@ -1,22 +0,0 @@
-# This file is maintained automatically by "terraform init".
```
erezrokah, line 1: See https://github.com/cloudquery/cq-provider-aws/pull/761

roneli (approved):

PR #795, opened 2022-04-26
chore: Add waf waf2 tf

zagronitay:


terraform/waf/modules/test/rule_group.tf
```diff
@@ -0,0 +1,32 @@
+resource "aws_waf_rule" "wafrule_2" {
```
erezrokah, line 1: [non-blocking] Should we follow the naming like other PRs and use `${var.prefix}-`? e.g. instead of `name = "${var.prefix}wafrule2"` use `name = "${var.prefix}-wafrule2"`

If this is due to a constraint on the name (it can't have `-` in it), please ignore

erezrokah (approved):
@zagronitay this looks great and thank you for following up on adding the prefix vars to other files.

I made a small non-blocking comment

terraform/waf/modules/test/rule_group.tf
zagronitay: Because this specific resource does not support non alphanumeric chars

roneli (approved):
LGTM

PR #797, opened 2022-04-26
fix: Adjust redshift PKs

### LINKED ISSUE #726: aws_redshift_snapshots PK Is not Unique
### Describe the bug

PK definition is not a unique identifier across all accounts

### Expected Behavior

PK definition should be unique in order to allow for insertion

### Steps to Reproduce

Create cluster with same name in different regions

### Possible Solution

Generate an ARN field then use that instead of the identifier which is only unique in a specific region

all

### Additional Context

_No response_

irmatov:
Closes #726


amanenk (approved):
LGTM

bbernays (approved):

PR #807, opened 2022-04-28
feat: Add retry log

disq:
Cleaned up

### DIFF
--- a/client/retry.go
+++ b/client/retry.go
@@ -18,17 +18,9 @@ type retryer struct {
 func newRetryer(logger hclog.Logger, maxRetries int, maxBackoff int) func() aws.Retryer {
 	return func() aws.Retryer {
 		return &retryer{
-			//Retryer: retry.NewStandard(func(o *retry.StandardOptions) {
-			//	o.MaxAttempts = maxRetries
-			//	o.MaxBackoff = time.Second * time.Duration(maxBackoff)
-			//}),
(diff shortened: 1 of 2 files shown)


client/retry.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+}
+
+func (r *retryer) RetryDelay(attempt int, err error) (time.Duration, error) {
+	dur, err := r.backoff.BackoffDelay(attempt, err)
+	if err != nil {
+		r.logger.Debug("retryDelay returned err", "err", err, "duration", dur.String())
+	} else {
+		r.logger.Debug("retryDelay will wait", "duration", dur.String())
+	}
```
roneli, line 40: extract resource type from error

roneli (changes requested):

client/retry.go
```diff
@@ -0,0 +1,42 @@
...
+				o.MaxBackoff = maxbackoff
+				o.Backoff = backoff
+			}),
+			backoff: backoff,
+			logger:  logger,
+		}
+	}
+}
+
+func (r *retryer) RetryDelay(attempt int, err error) (time.Duration, error) {
```
irmatov, line 34: Why not call Standard.RetryDelay here?

disq: Fixed 🙈

client/retry.go
disq: I tried

roneli (approved):
LGTM

PR #808, opened 2022-04-28
chore: Ensure added migration files version is correct

erezrokah:
Testing.

This ensures that if someone adds a migration, the order is an increment of the last migration, and the patch is an increment of the latest release


erezrokah:
This works only for patch increments. If we have breaking changes we increment the minor version, decided at release time.
So until we auto increment versions we can't validate the migration files on PRs.

Closing and will keep the branch for reference

PR #811, opened 2022-04-29
feat: Perform ec2 images calls concurrently

### LINKED ISSUE #767: EC2 AMI Performance Improvements
Currently we make 2 API calls to the Describe Images Endpoint. One has `Owners=self` and the other has `ExecutableUsers=self`. In large accounts these API calls can take a none trivial amount of time. We should investigate parallelizing the calls or combining them into the same API call. 

According to the AWS docs it is unclear whether adding multiple arguments to the API call treats them as an `and` or an `or`

irmatov:
Closes #767


disq (approved):
LGTM

PR #815, opened 2022-05-02
fix: Fetch backup tags only for the resources that support it

### LINKED ISSUE #725: Not All Backup Resources Support Tagging
### Describe the bug

From AWS CLI [Docs](https://docs.aws.amazon.com/cli/latest/reference/backup/list-tags.html)

Right now we assume all resources support tags

`
ListTags only works for resource types that support full Backup management of their backups. Those resource types are listed in the "Full Backup management" section of the [Feature availability by resource](https://docs.aws.amazon.com/aws-backup/latest/devguide/whatisbackup.html#features-by-resource) table.
`

### Expected Behavior

Should grab all appropriate tags

### Steps to Reproduce

Run fetch on a DynamoDB Backup Resource (`aws_backup_vault_recovery_points`)

### Possible Solution

1) Based on ARN of resource we need to decide if it supports tagging and if we decide it does then make the API call.

2) we could just ignore the error if it is a specific error that says Tagging is not supported

all

### Additional Context

_No response_

irmatov:
Closes #725


resources/services/backup/vaults.go
```diff
@@ -365,13 +367,37 @@ func resolveVaultRecoveryPointCreatedBy(ctx context.Context, meta schema.ClientM
...
+		return nil
+	}
+	resourceARN, err := arn.Parse(*rp.ResourceArn)
+	if err != nil {
+		return diag.WrapError(err)
+	}
+
+	// decide if the backed up resource supports tags
+	switch client.AWSService(resourceARN.Service) {
+	case client.S3Service, client.EFSService, client.DynamoDBService:
```
disq, line 380: empty case here to allow tags might raise eyebrows on reviews, maybe a comment?

disq (approved):
LGTM

bbernays (approved):

resources/services/backup/vaults.go
irmatov: :shrug: added a comment

PR #820, opened 2022-05-02
chore: Added Athena tf infrastructure

amanenk:
added terraform infrastructure for athena resources

### DIFF
--- a/terraform/athena/local/main.tf
+++ b/terraform/athena/local/main.tf
@@ -1,0 +1,4 @@
+module "demo" {
+  source = "../modules/test"
+  prefix = "cq"
+}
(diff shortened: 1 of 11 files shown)


terraform/athena/local/main.tf
```diff
@@ -0,0 +1,4 @@
+module "demo" {
+  source = "../modules/test"
+  prefix = "cq"
```
erezrokah, line 3:
```suggestion
  prefix = var.prefix
```

erezrokah (changes requested):
Looks good @amanenk. Left a small comment.

bbernays (approved):

erezrokah (approved):

PR #821, opened 2022-05-02
fix: Normalize InvalidSnapshot.NotFound error

erezrokah:

#### Summary

Related to https://github.com/cloudquery/cloudquery-issues/issues/327 (internal issue)

This PR redacts the snapshot id in the error
---

- [x] Read the [contribution guidelines](../CONTRIBUTING.md) 🧑‍🎓

### DIFF
--- a/client/errors.go
+++ b/client/errors.go
@@ -166,7 +166,7 @@ var (
 	encAuthRegex   = regexp.MustCompile(`(\s)(Encoded authorization failure message:)\s[A-Za-z0-9_-]+`)
 	userRegex      = regexp.MustCompile(`(\s)(is not authorized to perform: .+ on resource:\s)(user)\s.+`)
 	s3Regex        = regexp.MustCompile(`(\s)(S3(Key|Bucket))=(.+?)([,;\s])`)
-	snapshotRegex  = regexp.MustCompile(`(\s)(The snapshot 'snap-)(.+?)'( does not exist)`)
+	snapshotRegex  = regexp.MustCompile(`(\sThe )(.+ )'(.+?)'( does not exist)`)
 )
 
 func removePII(aa []Account, msg string) string {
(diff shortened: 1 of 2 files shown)


client/errors.go
```diff
@@ -166,6 +166,7 @@ var (
 	encAuthRegex   = regexp.MustCompile(`(\s)(Encoded authorization failure message:)\s[A-Za-z0-9_-]+`)
 	userRegex      = regexp.MustCompile(`(\s)(is not authorized to perform: .+ on resource:\s)(user)\s.+`)
 	s3Regex        = regexp.MustCompile(`(\s)(S3(Key|Bucket))=(.+?)([,;\s])`)
+	snapshotRegex  = regexp.MustCompile(`(\s)(The snapshot 'snap-)(.+?)'( does not exist)`)
```
roneli, line 169: I think maybe ` snapshot 'snap-` all of this can also be part of the regex as we can probably get similar errors of 
The instance i-adsa12321 does not exist.

Wdyt?

erezrokah: Sure that makes sense. Let me update it

erezrokah: Please let me know if https://github.com/cloudquery/cq-provider-aws/pull/821/commits/c7e4259e1f163eec85f0b4debcf283b76281cc13 is what you had in mind. I made the regex more generic, but still kept the resource type (so we don't group instance+snapshot)

roneli (approved):
LGTM

client/errors.go
roneli: Sounds great, yeah we should group same resoruces

client/errors.go
```diff
@@ -166,6 +166,7 @@ var (
 	encAuthRegex   = regexp.MustCompile(`(\s)(Encoded authorization failure message:)\s[A-Za-z0-9_-]+`)
 	userRegex      = regexp.MustCompile(`(\s)(is not authorized to perform: .+ on resource:\s)(user)\s.+`)
 	s3Regex        = regexp.MustCompile(`(\s)(S3(Key|Bucket))=(.+?)([,;\s])`)
+	snapshotRegex  = regexp.MustCompile(`(\sThe )(.+ )'(.+?)'( does not exist)`)
```
disq, line 169: so it's not "snapshot" any more and this is very aggressive. and already merged. @roneli

erezrokah: @disq This change is a follow up on https://github.com/cloudquery/cq-provider-aws/pull/821#discussion_r863490394

disq: @erezrokah that's the reason of the mention in my comment. it's still very aggressive and could gobble up whole messages as long as they fit the pattern.

erezrokah: I'm happy to revert that change pending @roneli's response

disq: Need a smaller regex I think. Maybe instead of `(.+ )` do something like `([A-Za-z0-9- ]+) `

erezrokah: Great suggestions. Done in https://github.com/cloudquery/cq-provider-aws/pull/827

PR #829, opened 2022-05-03
fix: Fix panic on fetching wafv2 without us-east-1

### LINKED ISSUE #828: fetching wafv2 resources causes panic when us-east-1 region is not enabled
### Describe the bug

If a user has an explicit list of regions that does not include us-east-1 and attempts to fetch wafv2 resources then fetch will panic will "nil pointer dereference"

### Expected Behavior

Fetch should succeed.

### Steps to Reproduce

Specify "us-east-2" as only item in regions. Do a fetch for wafv2 resources.

### Possible Solution

_No response_

latest core/aws

### Additional Context

_No response_

irmatov:
The bug happens when:

* user has a list of regions that doesn't include us-east-1
* wafv2 resources are fetched

The issue is that multiplexer always produces a client that has us-east-1 as a region and cloudfront scope.
When fetching starts, (*Client).Services call is made, which returns nil because there is no client for us-east-1 region.
This fix changes multiplexer to fetch cloudfront resources only when there is us-east-1 region.

Closes #828

### DIFF
--- a/client/multiplexers.go
+++ b/client/multiplexers.go
@@ -60,7 +60,7 @@ func ServiceAccountRegionScopeMultiplexer(service string) func(meta schema.Clien
 		client := meta.(*Client)
 		for accountID := range client.ServicesManager.services {
 			// always fetch cloudfront related resources
-			l = append(l, client.withAccountIDRegionAndScope(accountID, "us-east-1", wafv2types.ScopeCloudfront))
+			l = append(l, client.withAccountIDRegionAndScope(accountID, cloudfrontScopeRegion, wafv2types.ScopeCloudfront))
 			for region := range client.ServicesManager.services[accountID] {
 				if !isSupportedServiceForRegion(service, region) {
 					meta.Logger().Trace("region is not supported for service", "service", service, "region", region)
--- a/client/client.go
+++ b/client/client.go
@@ -91,6 +91,7 @@ const (
 	defaultRegion              = "us-east-1"
 	awsFailedToConfigureErrMsg = "failed to retrieve credentials for account %s. AWS Error: %w, detected aws env variables: %s"
 	defaultVar                 = "default"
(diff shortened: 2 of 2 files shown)


roneli (approved):
I guess this okay seems a little like a hack

client/multiplexers.go
```diff
@@ -59,13 +59,15 @@ func ServiceAccountRegionScopeMultiplexer(service string) func(meta schema.Clien
...
-			l = append(l, client.withAccountIDRegionAndScope(accountID, "us-east-1", wafv2types.ScopeCloudfront))
 			for region := range client.ServicesManager.services[accountID] {
 				if !isSupportedServiceForRegion(service, region) {
 					meta.Logger().Trace("region is not supported for service", "service", service, "region", region)
 					continue
 				}
+				if region == "us-east-1" {
+					// fetch cloudfront related resources only when us-east-1 is enabled
+					l = append(l, client.withAccountIDRegionAndScope(accountID, "us-east-1", wafv2types.ScopeCloudfront))
+				}
```
bbernays, line 70: Its not possible to disable `us-east-1` it is only possible to disable regions brought online after March 20, 2019:

```
Regions introduced before March 20, 2019 are enabled by default. You can begin creating and managing resources in these Regions immediately. You cannot enable or disable a Region that is enabled by default.
```

irmatov: "selected" is probably a better word. I meant like "enabled for fetching".

bbernays: I would argue that if a customer wants to grab WAF resources that should outweigh their specification of what regions to grab from. This is the same behavior that S3 has

irmatov: Reverted to older (desired) behavior of always fetching cloudfront scope

spangenberg (approved):

disq (approved):

PR #840, opened 2022-05-04
feat: Add S3 Bucket fetch speed improvement

spangenberg:


disq (commented):
lgtm, but why not use a waitgroup and range on the (potentially unbuffered) channel? range would stop iterating when the channel is closed, and no more len(response) business

spangenberg:
> lgtm, but why not use a waitgroup and range on the (potentially unbuffered) channel? range would stop iterating when the channel is closed, and no more len(response) business

But the response channel is still needed for the diagnostics, don't see where the WaitGroup would help in that case.

disq:
> But the response channel is still needed for the diagnostics, don't see where the WaitGroup would help in that case.

Create the waitgroup/goroutines, send stuff to the channel, close the channel, run a range on the errs channel (will need buffered channel or another, single goroutine to read err results into diags...) and to make sure all workers exited, finally do `wg.Wait()`? Better flow and readability/maintainability imo.

resources/services/s3/buckets.go
```diff
@@ -451,27 +453,53 @@ func S3Buckets() *schema.Table {
...
+	close(buckets)
+	go func() {
+		wg.Wait()
+		close(errs)
+	}()
 
-	res <- wb
-	return nil
+	var diags diag.Diagnostics
+	for err = range errs {
```
disq, line 481: `for err := range errs`, then do `diags.Add(err)` below?

spangenberg: Thanks for catching this one, did too much changing around and missed that one.

resources/services/s3/buckets.go
```diff
@@ -451,27 +453,53 @@ func S3Buckets() *schema.Table {
...
+	for i := 0; i < fetchS3BucketsPoolSize; i++ {
+		wg.Add(1)
+		go fetchS3BucketsWorker(ctx, meta, buckets, errs, res, &wg)
 	}
+	for _, bucket := range response.Buckets {
+		buckets <- bucket
+	}
+	close(buckets)
+	go func() {
+		wg.Wait()
```
disq, line 476: I would include the `for response.Buckets` (and close buckets chan) in this gofunc as well, otherwise (since we're using unbuffered channels) a deadlock is possible since we're not reading the errors channel (so if we don't have a worker for every bucket the send will fail

I would also check for early cancellation of context and switch those channel writes where to select statements where <-ctx.Done is also checked?

disq (approved):

resources/services/s3/buckets.go
```diff
@@ -451,27 +453,62 @@ func S3Buckets() *schema.Table {
...
+		go fetchS3BucketsWorker(ctx, meta, buckets, errs, res, &wg)
+	}
+	go func() {
+		defer close(buckets)
+		for _, bucket := range response.Buckets {
+			select {
+			case <-ctx.Done():
+				return
+			case buckets <- bucket:
+			}
```
roneli, line 478: very nice

roneli (approved):
LGTM

resources/services/s3/buckets.go
disq: no, super nice

PR #842, opened 2022-05-04
feat: Support AWS partitions

### LINKED ISSUE #806: Support China Region
### Describe the bug

When I fetch, it will report
```
unknown region: "cn-north-1"
```

### Expected Behavior

fetch success

### Steps to Reproduce

config.hcl
```
provider "aws" {
  configuration {

    regions = ["cn-north-1"]

  }
}
```

### Possible Solution

_No response_

Cloudquery: 0.22.10. AWS provider: 0.11.1

### Additional Context

_No response_

disq:
Should fix #806 (untested)

### DIFF
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -45,6 +45,18 @@ type SupportedServiceRegionsData struct {
 	Partitions map[string]AwsPartition `json:"partitions"`
 }
 
+func (s *SupportedServiceRegionsData) RegionsPartition(region string) (string, bool) {
+	for _, p := range s.Partitions {
+		for _, svc := range p.Services {
+			if _, ok := svc.Regions[region]; ok {
+				return p.Id, true
+			}
+		}
+	}
+
+	return defaultPartition, false
+}
+
 func readSupportedServiceRegions() *SupportedServiceRegionsData {
 	f, err := supportedServiceRegionFile.Open(PartitionServiceRegionFile)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


bbernays (approved):
LGTM

disq:
> Are there any tests we want to add for this?

added one, but not feeling very creative. Would need e2e to make sure I suppose

roneli (commented):
Ill comment as I am not 100% sure, but this means we run on all partitions now?
Should we add a config option to define what partitions we want to execute on? i.e by default don't have fetch from gov partitions, as I assume, most users don't have those regions enabled.
Other then that LGTM.

disq:
> Ill comment as I am not 100% sure, but this means we run on all partitions now? Should we add a config option to define what partitions we want to execute on? i.e by default don't have fetch from gov partitions, as I assume, most users don't have those regions enabled. Other then that LGTM.

We run on what describeregions return, or regions that are explicitly specified in the hcl. @roneli 
Shouldn't matter for normal users as describeregions only return aws-partition-regions for them.

roneli (approved):
LGTM

bbernays:
> > Ill comment as I am not 100% sure, but this means we run on all partitions now? Should we add a config option to define what partitions we want to execute on? i.e by default don't have fetch from gov partitions, as I assume, most users don't have those regions enabled. Other then that LGTM.
> 
> We run on what describeregions return, or regions that are explicitly specified in the hcl. @roneli Shouldn't matter for normal users as describeregions only return aws-partition-regions for them.

To add on to what @disq said: AWS credentials are scoped to a specific partition. So if a user uses a single provider block with multiple accounts in it they could still specify specific regions on a per account basis

PR #898, opened 2022-05-16
feat: Add partition info to regions

### LINKED ISSUE #416: Record and track AWS partition (aws, aws-cn, aws-us-gov)
### Describe the resource.

This probably doesn't need to be a first class `resource`, but AWS Partition should be tracked somwhere, like in `aws_accounts`.

### Use Case

At present, some resources don't include ARNs in their resource schema. With the current set materialized data, there is no way to create an ARN for those without making assumptions about the partition, or by grabbing the partition info from some type of resource that does materialize ARNs.

### Additional context

_No response_

irmatov:
Closes #416

### DIFF
--- a/resources/services/ec2/regions.go
+++ b/resources/services/ec2/regions.go
@@ -48,6 +48,12 @@ func AwsRegions() *schema.Table {
 				Type:        schema.TypeString,
 				Resolver:    schema.PathResolver("RegionName"),
 			},
+			{
+				Name:        "partition",
+				Description: "AWS partition",
+				Type:        schema.TypeString,
+				Resolver:    resolveRegionPartition,
+			},
 		},
 	}
 }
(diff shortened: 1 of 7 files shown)


resources/services/ec2/regions.go
```diff
@@ -78,3 +83,11 @@ func resolveRegionEnabled(ctx context.Context, meta schema.ClientMeta, resource
 	}
 	return nil
 }
+func resolveRegionPartition(ctx context.Context, meta schema.ClientMeta, resource *schema.Resource, c schema.Column) error {
+	region := resource.Item.(types.Region)
+	if region.RegionName == nil {
+		return nil
+	}
+	part, _ := client.RegionsPartition(*region.RegionName)
```
disq, line 91: I would check for `ok` and don't set if it was defaulted

disq (changes requested):

roneli (commented):
LGTM

roneli (approved):

disq (approved):

bbernays (changes requested):

PR #922, opened 2022-05-18
feat: Added Access Advisor for IAM 

### LINKED ISSUE #900: groups

### LINKED ISSUE #1389: IAM Access Advisor: groups

amanenk:

#### Summary

closes cloudquery/cloudquery#1389 

---

- [x] Lint your changes via `golangci-lint run --new-from-rev main` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))


resources/services/iam/groups_access_advisor.go
```diff
@@ -0,0 +1,188 @@
...
+func GroupsAccessAdvisors() *schema.Table {
+	return &schema.Table{
+		Name:          "aws_iam_groups_access_advisor",
+		Resolver:      fetchIamGroupsAccessAdvisors,
+		Multiplex:     client.AccountMultiplex,
+		IgnoreError:   client.IgnoreAccessDeniedServiceDisabled,
+		DeleteFilter:  client.DeleteAccountFilter,
+		IgnoreInTests: true,
+		Options:       schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "service_namespace"}},
+		Columns: []schema.Column{
```
bbernays, line 23: Will the table structure differ for IAM Roles/Policies/Users or could we make the table generic and just have specific resolvers for each type? I think that would make it easier for querying/using the data

amanenk: You are right. The structure is the same, but before listing access advisor resources we need to list users, groups, etc. So root fetch should be different. Some of the resolvers can be generic. I have added a commit that does it. Thanks for a hint.

resources/services/iam/groups_access_advisor.go
```diff
@@ -0,0 +1,178 @@
...
+//go:generate cq-gen --resource groups_access_advisor --config gen.hcl --output .
+func GroupsAccessAdvisors() *schema.Table {
+	return &schema.Table{
+		Name:          "aws_iam_groups_access_advisor",
+		Resolver:      fetchIamGroupsAccessAdvisors,
+		Multiplex:     client.AccountMultiplex,
+		IgnoreError:   client.IgnoreAccessDeniedServiceDisabled,
+		DeleteFilter:  client.DeleteAccountFilter,
+		IgnoreInTests: true,
+		Options:       schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "service_namespace"}},
```
bbernays, line 22: If we are going to use the same table for multiple entities then the PKs need something to differentiate between groups/roles and polices

bbernays: But if all of the data is the same except for a single column at the root table don't you think it would be simpler from a user perspective when querying?

bbernays: I can see tradeoffs on both sides of this. So I want to get others opinions

@roneli @zagronitay @yevgenypats  

We are working on implementing [4 resources](https://github.com/cloudquery/cq-provider-aws/issues/730) that have exactly the same fields. Do you think it is better to put the data into a single table and have a field in the root table to identify which resource type it is or have 4 different tables and infer the resource type by the table name?

resources/provider/migrations/postgres/32_v0.11.8.down.sql
```diff
@@ -0,0 +1,7 @@
+-- Autogenerated by migration tool on 2022-05-19 06:55:01
+-- CHANGEME: Verify or edit this file before proceeding
+
+-- Resource: iam.access_advisor_details
+DROP TABLE IF EXISTS aws_iam_access_advisor_tracked_actions_last_accessed;
```
bbernays, line 5: I think the migrations can be removed now

resources/services/iam/groups_access_advisor_details.go
```diff
@@ -0,0 +1,184 @@
...
+
+//go:generate cq-gen --resource groups_access_advisor_details --config gen.hcl --output .
+func GroupsAccessAdvisorDetails() *schema.Table {
+	return &schema.Table{
+		Name:          "aws_iam_access_advisor_details",
+		Description:   "IAM Access Advisor details of the IAM resources (users, groups, roles, policies)",
+		Resolver:      fetchIamGroupsAccessAdvisorDetails,
+		Multiplex:     client.AccountMultiplex,
+		IgnoreError:   client.IgnoreAccessDeniedServiceDisabled,
+		DeleteFilter:  AccessAdvisorFilter("group"),
```
bbernays, line 21:
```suggestion
		DeleteFilter:  AccessAdvisorFilter(GROUP),
```

resources/services/iam/roles_access_advisor_details.go
```diff
@@ -0,0 +1,184 @@
...
+		Resolver:      fetchIamRolesAccessAdvisorDetails,
+		Multiplex:     client.AccountMultiplex,
+		IgnoreError:   client.IgnoreAccessDeniedServiceDisabled,
+		DeleteFilter:  AccessAdvisorFilter(ROLE),
+		IgnoreInTests: true,
+		Options:       schema.TableCreationOptions{PrimaryKeys: []string{"account_id", "service_namespace"}},
+		Columns: []schema.Column{
+			{
+				Name:        "account_id",
+				Description: "The AWS Account ID of the resource.",
```
bbernays, line 27: Rather than duplicating all of the column definitions in each resource can we create a single place for them and then just reference it? This way it will be easier for people to understand that it is a single table and if they make changes in one file it will impact multiple resources

bbernays (commented):
Left a few comments, but overall when I run it on a clean install I get the following error:
```
Type: Internal Severity: Error
        Summary: table name aws_iam_access_advisor_detail_tracked_actions_last_accessed used more than once, duplicates are in iam.users_access_advisor_details and iam.roles_access_advisor_details
```

disq:
The duplicate name error on `aws_iam_access_advisor_detail_tracked_actions_last_accessed` comes from the fact that we're using the same table for multiple resources as results. It needs to be a separate name or an unrelational table that's defined only once. Having the same name isn't really handled anywhere so unless we overhaul the SDK and core to allow that, overriding the test wouldn't work as it'll just blow up somewhere else unexpectedly.

yevgenypats:
This needs removal of migrations and update to latest SDK

yevgenypats:
Should go to the monorepo https://github.com/cloudquery/cq-provider-aws/pull/922 if we still want to pursue it.

PR #1000, opened 2022-06-03
feat: Generic list and detail

### LINKED ISSUE #927: Create Reference Implementation Paginated List + Detail Fetch
Many resources require multiple API calls to fully resolve one of the most common patterns is a paginated list call with a nested for loop to get the resource details. This can be very expensive in terms of time required to fully resolve a resource.

The following resources have implemented some type of parallelization to improve this:
- `aws_sagemaker_training_jobs`
- `aws_ecs_task_definitions`
- `aws_athena_data_catalogs`

We should create a single reference implementation so that we can ensure we are catching all of the edge cases and also possibly implementing this in the SDK to reduce individual resource complexity

bbernays:

#### Summary

This pr Attempts to centralize the parallelization logic for resources in order to provide a singular location for tuning to occur (if necessary). While this initial implementation still has a hardcoded number of go routines, in the future we could swap out that hard coded number for a dynamic value based on resource utilization.

---

### DIFF
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -425,7 +425,6 @@ func ListAndDetailResolver(ctx context.Context, meta schema.ClientMeta, res chan
 	// All items will be attempted to be fetched, and all errors will be aggregated
 	<-done
 
-	// Only return the diags if there is an error
 	if diags.HasDiags() {
 		return diags
 	}
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -35,9 +35,9 @@ type SupportedServiceRegionsData struct {
 	regionVsPartition map[string]string
 }
 
-type ListResolver func(ctx context.Context, meta schema.ClientMeta) ([]interface{}, error)
+type ListResolverFunc func(ctx context.Context, meta schema.ClientMeta) ([]interface{}, error)
 
-type DetailResolver func(ctx context.Context, meta schema.ClientMeta, resultsChan chan<- interface{}, errorChan chan<- error, summary interface{})
+type DetailResolverFunc func(ctx context.Context, meta schema.ClientMeta, resultsChan chan<- interface{}, errorChan chan<- error, summary interface{})
 
 const (
 	ApigatewayService           AWSService = "apigateway"
@@ -386,11 +386,11 @@ func TagsToMap(tagSlice interface{}) map[string]string {
 	return ret
 }
 
-func ListAndDetailResolver(ctx context.Context, meta schema.ClientMeta, res chan<- interface{}, list ListResolver, details DetailResolver) error {
+func ListAndDetailResolver(ctx context.Context, meta schema.ClientMeta, res chan<- interface{}, list ListResolverFunc, details DetailResolverFunc) error {
 	var diags diag.Diagnostics
 	var wg sync.WaitGroup
 	errorChan := make(chan error)
-
+	// detailChan := make(chan interface{})
 	// Channel that will communicate with goroutine that is aggregating the errors
 	done := make(chan struct{})
 	go func() {
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -410,7 +410,6 @@ func ListAndDetailResolver(ctx context.Context, meta schema.ClientMeta, res chan
 			defer wg.Done()
 			defer sem.Release(1)
 			details(ctx, meta, res, errorChan, summary)
-
 		}(item)
 	}
 	// Ensure all items have been attempted to be fetched
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -420,9 +420,9 @@ func ListAndDetailResolver(ctx context.Context, meta schema.ClientMeta, res chan
 	if err != nil {
 		return diag.WrapError(err)
 	}
-	// All items will be attempted to be fetched, but could return an error
 
-	// This will trigger aggregating go routine to stop
+	// All items will be attempted to be fetched, and all errors will be aggregated
+	<-done
 
 	// Only return the diags if there is an error
 	if diags.HasErrors() {
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -424,7 +424,7 @@ func ListAndDetailResolver(ctx context.Context, meta schema.ClientMeta, res chan
 	<-done
 
 	// Only return the diags if there is an error
-	if diags.HasErrors() {
+	if diags.HasDiags() {
 		return diags
 	}
 	return nil
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -412,7 +412,6 @@ func ListAndDetailResolver(ctx context.Context, meta schema.ClientMeta, res chan
 				details(ctx, meta, res, errorChan, summary)
 			}(item)
 		}
-
 	}()
 
 	err := list(ctx, meta, detailChan)
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -13,7 +13,10 @@ import (
 
 	"github.com/aws/aws-sdk-go-v2/aws/arn"
 	"github.com/aws/smithy-go"
+	"github.com/cloudquery/cq-provider-sdk/provider/diag"
 	"github.com/cloudquery/cq-provider-sdk/provider/schema"
+	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/semaphore"
 )
 
 type AWSService string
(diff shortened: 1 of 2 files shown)


client/helpers.go
```diff
@@ -382,3 +391,28 @@ func TagsToMap(tagSlice interface{}) map[string]string {
 	TagsIntoMap(tagSlice, ret)
 	return ret
 }
+
+func ListAndDetailResolver(ctx context.Context, meta schema.ClientMeta, res chan<- interface{}, list ListResolver, details DetailResolver) error {
+	sem := semaphore.NewWeighted(int64(MAX_GOROUTINES))
+	errs, ctx := errgroup.WithContext(ctx)
+	response, err := list(ctx, meta)
```
roneli, line 398: what if the list is paginated?

client/helpers.go
```diff
@@ -382,3 +391,28 @@ func TagsToMap(tagSlice interface{}) map[string]string {
...
+		return diag.WrapError(err)
+	}
+	for _, item := range response {
+		if err := sem.Acquire(ctx, 1); err != nil {
+			return diag.WrapError(err)
+		}
+		func(summary interface{}) {
+			errs.Go(func() error {
+				defer sem.Release(1)
+				return details(ctx, meta, res, summary)
```
roneli, line 409: if one details fails they all fail as its an error group, perhaps aggregate all diags and return in the end?

roneli (commented):
Looking good, had a few comments.

client/helpers.go
bbernays: The `list` function is supposed to handle all of the pagination and only return when it is complete

client/helpers.go
bbernays: Good suggestion

client/helpers.go
```diff
@@ -382,3 +390,39 @@ func TagsToMap(tagSlice interface{}) map[string]string {
...
+			defer wg.Done()
+			defer sem.Release(1)
+			details(ctx, meta, res, errorChan, summary)
+		}(item)
+	}
+	// Ensure all items have been attempted to be fetched
+	wg.Wait()
+
+	// empty the error channel and convert to Diags
+	for len(errorChan) > 0 {
```
disq, line 419: why not do a range? you can `close(errorChan)` after `wg.Wait` returns.

disq: you'd need to use a buffered channel and preallocate (limited) "slots" for potential errors.

Also since errorChan is unbuffered you even need to start reading in a separate goroutine before firing off the first worker... then `range` will stop reading once the channel is closed. See https://github.com/cloudquery/cq-provider-sdk/pull/208/files for an example.

yevgenypats (dismissed):
Unless this is a blocker for a client I would hold off any non critical speed and new features for aws provider until we are able to stabilize it and get around bugs/sentry.

client/helpers.go
```diff
@@ -33,6 +35,10 @@ type SupportedServiceRegionsData struct {
 	regionVsPartition map[string]string
 }
 
+type ListResolver func(ctx context.Context, meta schema.ClientMeta) ([]interface{}, error)
```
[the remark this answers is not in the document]
disq: Another point, why not have the signature somewhat match DetailResolver? ie. List gets a channel to write summaries to, then we listen and route that (or maybe give the same channel and not do much routing) to Detail? That way we don't need to wait for the pagination/listing to finish before we start getting actual results.

disq:
This should ideally be refactored in the SDK maybe? not sure

client/helpers.go
disq: The changes look good 👍

bbernays:
> This should ideally be refactored in the SDK maybe? not sure

I think you are correct, but I am not trying to get ahead of ourselves. I am trying to be slow/deliberate going from resolver --> client --> sdk. 

I think I would only put this in the SDK if we implemented the same pattern in multiple providers

yevgenypats:
@bbernays can you please point me to the 3 places in the code we use this pattern?

bbernays:
> @bbernays can you please point me to the 3 places in the code we use this pattern?

https://github.com/cloudquery/cq-provider-aws/blob/53d391b821f2a1340ad04aa50a9a8fbe1744ac11/resources/services/athena/data_catalogs.go#L221-L249

https://github.com/cloudquery/cq-provider-aws/blob/53d391b821f2a1340ad04aa50a9a8fbe1744ac11/resources/services/athena/work_groups.go#L416-L442

https://github.com/cloudquery/cq-provider-aws/blob/53d391b821f2a1340ad04aa50a9a8fbe1744ac11/resources/services/sagemaker/training_jobs.go#L590-L617

https://github.com/cloudquery/cq-provider-aws/blob/53d391b821f2a1340ad04aa50a9a8fbe1744ac11/resources/services/ecs/task_definitions.go#L647-L678

client/helpers.go
roneli: but that could create a huge list of data and memory spikes, it worth noting this to the implementor

client/helpers.go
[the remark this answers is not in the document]
bbernays: I have updated the function names as suggested and added the channel to communicate between list and detail. I am not sure what you mean for GoDoc, can you point me in the direction of documentation or an example of what you are looking for?

client/helpers.go
```diff
@@ -377,3 +385,47 @@ func TagsToMap(tagSlice interface{}) map[string]string {
...
+	close(detailChan)
+	if err != nil {
+		return diag.WrapError(err)
+	}
+
+	// All items will be attempted to be fetched, and all errors will be aggregated
+	<-done
+
+	// Only return the diags if there is an error
+	if diags.HasErrors() {
```
disq, line 427:
```suggestion
	if diags.HasDiags() {
```
HasErrors implies ERROR severity but it could also return warnings

resources/services/athena/work_groups.go
```diff
@@ -582,21 +572,22 @@ func fetchAthenaWorkGroupNamedQueries(ctx context.Context, meta schema.ClientMet
...
 	}, func(options *athena.Options) {
 		options.Region = c.Region
 	})
 	if err != nil {
 		if c.IsNotFoundError(err) {
-			return nil
+			return
 		}
-		return diag.WrapError(err)
+		errorChan <- diag.WrapError(err)
```
disq, line 588: add `return` afterwards or it will panic on L590 if dc is `nil`

client/helpers.go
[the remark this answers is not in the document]
disq: things like `// ListResolver should do this and this and return this` etc.

client/helpers.go
bbernays: Now there will not be a spike as there is a channel between the list and detail

client/helpers.go
```diff
@@ -377,3 +387,47 @@ func TagsToMap(tagSlice interface{}) map[string]string {
...
+	err := list(ctx, meta, detailChan)
+	close(detailChan)
+	if err != nil {
+		return diag.WrapError(err)
+	}
+
+	// All items will be attempted to be fetched, and all errors will be aggregated
+	<-done
+
+	// Only return the diags if there is an error
```
disq, line 428: Outdated comment w.r.t. we're checking `diags.HasDiags()` below (as suggested)

resources/services/athena/work_groups.go
```diff
@@ -12,16 +12,17 @@ import (
...
 //go:generate cq-gen --resource work_groups --config gen.hcl --output .
 func WorkGroups() *schema.Table {
 	return &schema.Table{
-		Name:         "aws_athena_work_groups",
-		Description:  "A workgroup, which contains a name, description, creation time, state, and other configuration, listed under WorkGroup$Configuration",
-		Resolver:     fetchAthenaWorkGroups,
+		Name:        "aws_athena_work_groups",
+		Description: "A workgroup, which contains a name, description, creation time, state, and other configuration, listed under WorkGroup$Configuration",
+		Resolver: func(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
+			err := diag.WrapError(client.ListAndDetailResolver(ctx, meta, res, listWorkGroups, workGroupDetail))
```
disq, line 23: Could be one less line here (just `return diag.WrapError(...`)

disq (approved):
LGTM with couple nits

PR #1006, opened 2022-06-06
fix: Not Supported region added to error classifier

amanenk:

#### Summary

Some ec2 resources does not allows some operations and return error like: `operation error EC2: DescribeCustomerGateways, https response error StatusCode: 400, RequestID:xxxxxxxxxxxxxxxxxxxxxxxxx, api error UnsupportedOperation: The functionality you requested is not available in this region.`
The pr adds this kind of errors to classifier that ignores them.

---

- [x] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))

### DIFF
--- a/client/errors.go
+++ b/client/errors.go
@@ -139,7 +139,6 @@ func classifyError(err error, fallbackType diag.Type, accounts []Account, opts .
 				),
 			}
 		}
-
 	}
 	if IsErrorThrottle(err) {
 		return diag.Diagnostics{
--- a/client/errors.go
+++ b/client/errors.go
@@ -112,6 +112,19 @@ func classifyError(err error, fallbackType diag.Type, accounts []Account, opts .
 					)...),
 				),
 			}
+		case "UnsupportedOperation":
+			if strings.Contains(ae.ErrorMessage(), "The functionality you requested is not available in this region.") {
+				return diag.Diagnostics{
+					RedactError(accounts, diag.NewBaseError(err,
+						diag.RESOLVING,
+						append(opts,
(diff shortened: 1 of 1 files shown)


client/errors.go
```diff
@@ -112,6 +112,19 @@ func classifyError(err error, fallbackType diag.Type, accounts []Account, opts .
...
 				),
 			}
+		case "UnsupportedOperation":
+			if strings.Contains(ae.ErrorMessage(), "The functionality you requested is not available in this region.") {
+				return diag.Diagnostics{
+					RedactError(accounts, diag.NewBaseError(err,
+						diag.RESOLVING,
+						append(opts,
+							diag.WithType(diag.RESOLVING),
+							diag.WithSeverity(diag.IGNORE),
```
roneli, line 122: if we put this in ignore users won't see this. cc @bbernays

roneli (approved):
LGTM, have one conversation would like to resolve if severity should be warning or ignore. I think the latter.

client/errors.go
amanenk: I am not sure which type is better as well. My thoughts are next: the case we have here is not fixable by user so we don't need to notify them. For example if user fetches all regions they will always have warnings that they can't fix.

bbernays: I agree with @amanenk. If the user cannot fix the issue then why notify them. The only scenario I can see this not being the behavior we want is if the user specifies only a single region where the resource is not available. In that one case I would expect that the user would just remove the resource from the fetch list

bbernays (approved):

PR #1010, opened 2022-06-07
fix: Ignore 404 errors in apigateway rest apis

irmatov:


bbernays:
I don't think we need to handle this in the resolver code. The issue stems from the nested tables don't have IgnoreError handlers on them. @yevgenypats has a pr for making nested tables inherit the IgnoreError handler from the root table and I have a pr for explicitly adding a handler on all tables

PR #1019, opened 2022-06-08
fix: Panic in IsInvalidParameterValueError

irmatov:
Closes https://github.com/cloudquery/cloudquery-issues/issues/381

### DIFF
--- a/client/helpers.go
+++ b/client/helpers.go
@@ -302,11 +302,7 @@ func isNotFoundError(err error) bool {
 
 func IsInvalidParameterValueError(err error) bool {
 	var apiErr smithy.APIError
-	if errors.As(err, &apiErr); apiErr.ErrorCode() == "InvalidParameterValue" {
-		return true
-	}
-
-	return false
+	return errors.As(err, &apiErr) && apiErr.ErrorCode() == "InvalidParameterValue"
 }
 
 func IsAWSError(err error, code ...string) bool {


yevgenypats:
Does it fix it? why behaviour is different with explicit variable here?

client/helpers.go
```diff
@@ -302,7 +302,7 @@ func isNotFoundError(err error) bool {
 
 func IsInvalidParameterValueError(err error) bool {
 	var apiErr smithy.APIError
-	if errors.As(err, &apiErr); apiErr.ErrorCode() == "InvalidParameterValue" {
+	if ok := errors.As(err, &apiErr); ok && apiErr.ErrorCode() == "InvalidParameterValue" {
```
disq, line 305:
```suggestion
	if errors.As(err, &apiErr) && apiErr.ErrorCode() == "InvalidParameterValue" {
```
simpler?

yevgenypats:
Looks like the original way is preferable per their docs - https://pkg.go.dev/errors

can you elaborate or add a test to make sure it fix it?

client/helpers.go
yevgenypats: yes this should be the way to go and the suggested one

irmatov: @yevgenypats You're missing that current code does not test a result of `errors.As`, so if it is `false` then `apiErr` is nil thus panic.

disq (approved):

erezrokah:
Thanks for the fix @irmatov and the review @disq and @yevgenypats

PR #1030, opened 2022-06-10
fix: Continue fetching on incorrect account permissions

### LINKED ISSUE #1024: incorrect message in org feature
### Describe the bug

When using org feature of aws with the following config

```
   org {
      member_role_name = "OrganizationAccountAccessRole"
    }
```

If I don't have access I'll run `cloudquery fetch` and will get the following message:

```

Type: Internal Severity: Error
        Summary: rpc error: code = Unknown desc = provider client is not configured (Hint: Try upgrading cloudquery)

Provider aws fetch summary: ✓ Total Resources fetched: 0         ⚠️ Warnings: 0        ❌ Errors: 0
```

### Expected Behavior

Getting the right error message with authentication denied or something like that

### Steps to Reproduce

-

### Possible Solution

_No response_

v0.12.7

### Additional Context

_No response_

### LINKED ISSUE #1026: Multiple Accounts with incorrect permissions are stopping fetch
### Describe the bug

Re posting discord issue:

When a user defines multiple accounts but one of them has incorrect permissions it will still fail the whole fetch.

### Expected Behavior

continue fetching.

### Steps to Reproduce

-

### Possible Solution

_No response_

-

### Additional Context

_No response_

### LINKED ISSUE #1031: Specifying Different Regions for Different Accounts Results in a panic
### Describe the bug

If a user specifies different regions for different accounts then CQ can panic. Here is an example config that will crash:

```
provider "aws" {
 configuration {
    accounts "account1" {
      local_profile = "account-1"
      regions = ["us-west-2"]  
    }   

    accounts "account2" {
      regions = ["us-east-2"]
      local_profile = "account-2"
    }

  }
  resources = ["iam.accounts"]

}```

### Expected Behavior

Should fetch successfully

### Steps to Reproduce

see above

### Possible Solution

_No response_

all versions

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by bbernays**

#### Summary

---

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -73,6 +73,7 @@ import (
 	wafv2types "github.com/aws/aws-sdk-go-v2/service/wafv2/types"
 	"github.com/aws/aws-sdk-go-v2/service/workspaces"
 	"github.com/aws/aws-sdk-go-v2/service/xray"
+	"github.com/aws/smithy-go"
 	"github.com/aws/smithy-go/logging"
 	"github.com/cloudquery/cq-provider-sdk/provider/diag"
 	"github.com/cloudquery/cq-provider-sdk/provider/schema"
@@ -82,7 +83,6 @@ import (
 type Client struct {
 	// Those are already normalized values after configure and this is why we don't want to hold
 	// config directly.
-	Accounts        []Account
 	logLevel        *string
 	maxRetries      int
 	maxBackoff      int
@@ -174,12 +174,12 @@ type Services struct {
 	Xray                   XrayClient
 }
 
-type ServicesAccountRegionMap map[string]map[string]*Services
+type ServicesPartitionAccountRegionMap map[string]map[string]map[string]*Services
 
 // ServicesManager will hold the entire map of (account X region) services
 type ServicesManager struct {
-	services         ServicesAccountRegionMap
-	wafScopeServices map[string]*Services
+	services         ServicesPartitionAccountRegionMap
+	wafScopeServices map[string]map[string]*Services
 }
 
 const (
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by yevgenypats — APPROVED** [review]
Looks good.  This code section in general might need additional refactor in the future but for the sake of unblocking this and fixing number of issues that block customers we can push this now. 

Just some idea - remove `InitServicesForPartitionAccountAndScope` as we only need to create aws clients per region per account rather per some parameter - Otherwise we are not alligned with AWS server side throttling mechanism and we will get throttle more.

**Review by disq — COMMENTED** [review]
Why the partition business? A credential only be valid in a single partition, no? So why not have the partition as a property in `Client` and get rid of the additional map levels?

**Comment by bbernays** [comment]
> Why the partition business? A credential only be valid in a single partition, no? So why not have the partition as a property in `Client` and get rid of the additional map levels?

@disq - We need to store the partition in that serviceregion map because the `client.partition` field can be overwritten if there are credentials from different partitions

**Comment by bbernays** [comment]
> Looks good. This code section in general might need additional refactor in the future but for the sake of unblocking this and fixing number of issues that block customers we can push this now.
> 
> Just some idea - remove `InitServicesForPartitionAccountAndScope` as we only need to create aws clients per region per account rather per some parameter - Otherwise we are not alligned with AWS server side throttling mechanism and we will get throttle more.

@yevgenypats - Completely agree that more code can be refactored, but was trying to balance changing code with fixing issues. I would not delete `InitServicesForPartitionAccountAndScope` in this pr as I would want to take a deeper look into that change

**Review by disq — APPROVED** [review]
~Looks like `Account` struct can be removed now,~ LGTM ~otherwise~

File: client/client.go [hunk 1]
```diff
@@ -244,17 +255,26 @@ func (s3Manager S3Manager) GetBucketRegion(ctx context.Context, bucket string, o
...
+}
+
+func (s ServicesPartitionAccountRegionMap) Accounts() []string {
+	accounts := make([]string, 0)
+	for partitions := range s {
+		for account := range s[partitions] {
+			accounts = append(accounts, account)
+		}
 	}
+	return accounts
```
irmatov, line 274: do we care about duplicates here?

bbernays: I don't think an AWS account can span multiple partitions

PR #1033, opened 2022-06-10
fix: Change from IgnoreAccessDeniedServiceDisabled to IgnoreCommonErrors

bbernays:

#### Summary

When we use `IgnoreAccessDeniedServiceDisabled` we have to handle certain errors like 404/ Resource not found errors in the resolvers. By changing to IgnoreCommonErrors we get that at the table level without any extra code.

I am not 100% sure this is valid or if there are cases when we do NOT want to ignore 404s

---

### DIFF
--- a/resources/services/accessanalyzer/analyzers.go
+++ b/resources/services/accessanalyzer/analyzers.go
@@ -20,7 +20,7 @@ func Analyzers() *schema.Table {
 		Description:  "Contains information about the analyzer",
 		Resolver:     fetchAccessAnalyzerAnalyzers,
 		Multiplex:    client.ServiceAccountRegionMultiplexer("access-analyzer"),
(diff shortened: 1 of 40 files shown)


erezrokah (approved):

roneli (approved):
LGTM

Regarding the edge cases, in general the resolver still needs to handle not founds specially in cases where its listing and doing on each on an action.

Root resolver: resolves 100 resources
 - relation resolver: for each resource resolve -  if not found (with new ignore error + using root table ignore error, it will just not resolve and not be inserted in DB), same with columns

On the other hand on Root Resolver if we do this logic and return on a not found we won't resolve the next 100 resources, so it should continue to resolve if it gets a not found.

disq:
I was clear for me but now I'm all confused 😂

PR #1056, opened 2022-06-14
fix: Classify credential errors as USER type

irmatov:
Prevents sending

    : failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds:
    GetMetadata, request canceled, context deadline exceeded

error handling

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -446,7 +446,12 @@ func configureAwsClient(ctx context.Context, logger hclog.Logger, awsConfig *Con
 	// Test out retrieving credentials
 	if _, err := awsCfg.Credentials.Retrieve(ctx); err != nil {
 		logger.Error("error retrieving credentials", "err", err)
-		return awsCfg, err
+		return awsCfg, diag.FromError(
+			err,
+			diag.USER,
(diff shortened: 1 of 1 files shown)


bbernays (approved):

client/client.go
```diff
@@ -446,7 +446,7 @@ func configureAwsClient(ctx context.Context, logger hclog.Logger, awsConfig *Con
 	// Test out retrieving credentials
 	if _, err := awsCfg.Credentials.Retrieve(ctx); err != nil {
 		logger.Error("error retrieving credentials", "err", err)
-		return awsCfg, err
+		return awsCfg, diag.FromError(err, diag.USER)
```
bbernays, line 449: I think we should actually be adding a message to this error to tell users that credentials were not found.

We are seeing this error here because the SDK has gone through all of the places it looks for credentials and hasn't found anything.

Here are docs on where it [looks](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html#cli-configure-quickstart-precedence)

irmatov: Thanks, updated. Take a look, please.

PR #1059, opened 2022-06-15
fix: Resolvers Returning Early

### LINKED ISSUE #1058: Lambda Function Ending Early
In certain cases when a function is not found no other functions will be resolved

https://github.com/cloudquery/cq-provider-aws/blob/668ea91620f85994606935d3b7a8f171a7d8e8a7/resources/services/lambda/functions.go#L1110-L1115

bbernays:

#### Summary

---

### DIFF
--- a/resources/services/dynamodb/tables.go
+++ b/resources/services/dynamodb/tables.go
@@ -524,7 +524,7 @@ func fetchDynamodbTables(ctx context.Context, meta schema.ClientMeta, parent *sc
 			})
 			if err != nil {
 				if c.IsNotFoundError(err) {
-					return nil
+					continue
 				}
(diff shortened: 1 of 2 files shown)


yevgenypats (changes requested):
Looks good but  I think we should then add nil checks in subsequent resolvers otherwise it will panic on those?

bbernays:
> Looks good but I think we should then add nil checks in subsequent resolvers otherwise it will panic on those?

Why? we are only sending the data to the `res` channel if there are no errors. These are both examples of where we do a list then a fetch for detail

yevgenypats (approved):
Yeah you are right I thought it was a subresovler, in that case it should work

PR #1071, opened 2022-06-20
fix: Redact separate error messages separately

irmatov:
This splits single lookupRegex into two:

* one matches lookup failures
* second matches read errors


disq (approved):

erezrokah (approved):

irmatov:
Thanks for such a swift review @disq @erezrokah !

PR #1092, opened 2022-06-22
feat: Add ecr image scan results

### LINKED ISSUE #637: Support Enhanced Image Scanning Results
### Describe the resource.

Enhanced Image Scanning results should be retrieved by CQ

### Use Case

Understanding vulnerabilities of ecr images

### Additional context

_No response_

### LINKED ISSUE #1408: Support Enhanced Image Scanning Results
### Describe the resource.

Enhanced Image Scanning results should be retrieved by CQ

### Use Case

Understanding vulnerabilities of ecr images

### Additional context

_No response_

irmatov:
Closes cloudquery/cloudquery#1408

### DIFF
--- a/resources/services/ecr/repositories.go
+++ b/resources/services/ecr/repositories.go
@@ -2,6 +2,7 @@ package ecr
 
 import (
 	"context"
+	"encoding/json"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/ecr"
@@ -10,6 +11,8 @@ import (
 	"github.com/cloudquery/cq-provider-sdk/provider/diag"
 	"github.com/cloudquery/cq-provider-sdk/provider/schema"
 )
+
+//go:generate cq-gen --resource repositories --config repositories.hcl --output .
 
 func Repositories() *schema.Table {
 	return &schema.Table{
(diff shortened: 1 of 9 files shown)


resources/services/ecr/repositories.go
```diff
@@ -241,3 +563,78 @@ func fetchEcrRepositoryImages(ctx context.Context, meta schema.ClientMeta, paren
...
+		ImageId:        &types.ImageIdentifier{ImageDigest: image.ImageDigest},
+		RepositoryName: image.RepositoryName,
+		MaxResults:     &maxResults,
+	}
+	c := meta.(*client.Client)
+	svc := c.Services().ECR
+	for {
+		output, err := svc.DescribeImageScanFindings(ctx, &config, func(options *ecr.Options) {
+			options.Region = c.Region
+		})
```
bbernays, line 579: This is the call I am worried about, Every Repo will have to make at least `1` API call for each image, and each of those extra calls is done sequentially

hermanschaaf: @bbernays @irmatov  I see this PR has been parked for a while now--is this the blocker, and is there an alternative solution we can consider?

bbernays: The only other solution is to make `image scan results` be its own table. In order to do that we would have to duplicate the call to list repositories call, which might be a reasonable trade off...

candiduslynx (approved):

PR #1100, opened 2022-06-23
fix: Improve Errors AWS Errors

bbernays:
#### Summary

- Improve Error message when CQ is unable to assume a role in a org member account
- Enable Account and Org to be in same config
---

- [x] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -465,11 +465,6 @@ func Configure(logger hclog.Logger, providerConfig interface{}) (schema.ClientMe
 	client := NewAwsClient(logger)
 	client.GlobalRegion = awsConfig.GlobalRegion
 	var adminAccountSts AssumeRoleAPIClient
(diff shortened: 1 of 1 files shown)


yevgenypats (changes requested):
Looks good but I think we need to introduce tests for that both to validate this is correct and also to help with future regression and maybe add docs supporting that

bbernays:
> Looks good but I think we need to introduce tests for that both to validate this is correct and also to help with future regression and maybe add docs supporting that

Great catch. There would have been regressions had this merged with how the STS client is passed around to assume roles. I have updated the PR to explicitly return an error if a user tries to create this specification until we can do a larger refactor.

yevgenypats (approved):

PR #1167, opened 2022-07-05
chore(dev): Add pre-commit hook documentation

hermanschaaf:
This adds documentation for the option of installing a pre-commit hook that runs `go mod tidy` and `golangci-lint`. I decided to keep it fairly lean for now; we can add more checks as they come up.


hermanschaaf:
Update: the feature itself was now added via an automated sync from the cloudquery/.github repo in https://github.com/cloudquery/cq-provider-aws/pull/1168 - this change just updates the documentation.

yevgenypats (approved):

PR #1172, opened 2022-07-06
fix: Ignore s3 "bucket not found" error

irmatov:
Should fix weird error:


resources/services/s3/buckets.go
```diff
@@ -531,9 +531,6 @@ func resolveS3BucketsAttributes(ctx context.Context, meta schema.ClientMeta, res
 
 	output, err := mgr.GetBucketRegion(ctx, *resource.Name)
 	if err != nil {
-		if c.IsNotFoundError(err) {
```
irmatov, line 534: this should be correctly handled higher up the stack

irmatov: I reverted the change

disq (approved):

PR #1190, opened 2022-07-08
fix: Classify DNS errors as user

irmatov:


client/errors.go
```diff
@@ -182,6 +183,19 @@ func classifyError(err error, fallbackType diag.Type, accounts []string, opts ..
...
 	}
+	if dnsErrorRegex.MatchString(err.Error()) {
+		return diag.Diagnostics{
+			RedactError(accounts, diag.NewBaseError(err,
+				diag.USER,
+				append(opts,
+					diag.WithType(diag.USER),
+					diag.WithSeverity(diag.ERROR),
+					ParseSummaryMessage(err),
+					diag.WithDetails("Encountered a DNS error. Please check your DNS and networking settings and try again."),
```
bbernays, line 194: What should a user be checking for?

disq: I think it's ok

disq (approved):

PR #1194, opened 2022-07-08
feat: Partial lambda Function  Fetch

### LINKED ISSUE #1191: Lambda table not been populated when using ViewOnlyAccess
### Describe the bug

Although "Lambda:List*" is present, no error is received and Lambda tables are completely empty when running Cloudquery with ViewOnlyAccess set of permissions.

### Expected Behavior

Populate the tables with the data we can get using the "lambda:List*" permissions present on the ViewOnlyAccess

### Steps to Reproduce

Execute Cloudquery with ViewOnlyAccess set of permissions.

### Possible Solution

_No response_

AWS

### Additional Context

_No response_

bbernays:
#### Summary

Here is an experimental change, where we can populate some data for Lambda Functions as long as the user has access to `lambda:List*` 

Pros:
- As long as user has access to List they can get some data in the table, they will miss information `types.FunctionCodeLocation`, `types.Concurrency` and `Tags`

Cons:
- Includes information about function that could have been deleted
- Continues to try to get details as we cannot be sure if a user doesn't have access to get a specific function or all functions

---


client/helpers.go
```diff
@@ -301,6 +307,29 @@ func isNotFoundError(err error) bool {
 	return false
 }
 
+// IsAccessDeniedError checks if api error should be classified as a permissions issue
+func (c *Client) IsAccessDeniedError(err error) bool {
+	if isAccessDeniedError(err) {
+		c.logger.Warn("API returned \"NotFound\" error ignoring it...", "error", err)
```
candiduslynx, line 313: NotFound -> AccessDenied

client/helpers.go
```diff
@@ -301,6 +307,29 @@ func isNotFoundError(err error) bool {
...
+	return false
+}
+
+func isAccessDeniedError(err error) bool {
+	var ae smithy.APIError
+	if !errors.As(err, &ae) {
+		return false
+	}
+	errorCode := ae.ErrorCode()
+	for _, s := range accessDeniedErrorStrings {
```
candiduslynx, line 325: maybe do a map instead?
```go
var accessDeniedErrorStrings = map[string]struct{}{
	"AuthorizationError": struct{}{},
	"AccessDenied": struct{}{},
	"AccessDeniedException": struct{}{},
	"InsufficientPrivilegesException": struct{}{},
	"UnauthorizedOperation": struct{}{},
}

func isAccessDeniedError(err error) bool {
	var ae smithy.APIError
	if !errors.As(err, &ae) {
		return false
	}
	errorCode := ae.ErrorCode()
	_, ok := accessDeniedErrorStrings[errorCode]
	return ok
}
```

candiduslynx (changes requested):

disq (approved):

client/helpers.go
```diff
@@ -301,6 +307,24 @@ func isNotFoundError(err error) bool {
 	return false
 }
 
+// IsAccessDeniedError checks if api error should be classified as a permissions issue
+func (c *Client) IsAccessDeniedError(err error) bool {
+	if isAccessDeniedError(err) {
+		c.logger.Warn("API returned \"Access Denied\" error ignoring it...", "error", err)
```
candiduslynx, line 313: Access Denied -> AccessDenied ?

candiduslynx (approved):

client/helpers.go
disq: Actually we don't know if it's `Access Denied` or some other error in the `accessDeniedErrorStrings` map, so log line not that accurate.

yevgenypats (approved):

client/helpers.go
bbernays: Updated the message

PR #1220, opened 2022-07-14
feat: Policies

disq:
Adds SQL policies to the provider. We will follow up with more details soon.


.github/workflows/test_policy_sql.yml
```diff
@@ -0,0 +1,83 @@
...
+          --health-timeout 5s
+          --health-retries 5
+    steps:
+      - name: Check out code into the policy directory
+        uses: actions/checkout@v3
+
+      - name: Download Cloudquery
+        run: |
+          curl -L https://github.com/cloudquery/cloudquery/releases/latest/download/cloudquery_Linux_x86_64 -o cloudquery
+          chmod a+x cloudquery
```
bbernays, line 45: I think we should be using the Github Action that we have on the marketplace

.github/workflows/test_policy_sql.yml
```diff
@@ -0,0 +1,83 @@
...
+        run: |
+          set -o xtrace
+
+          go run ./main.go &
+
+          # Wait until .cq_reattach exists
+          bash -c "until test -f .cq_reattach; do echo 123; sleep 1; done"
+
+          export CQ_REATTACH_PROVIDERS=.cq_reattach
+          ./cloudquery provider sync aws --config ./test/policy_cq_config.yml --enable-console-log
```
yevgenypats, line 79: I would do it in two steps as will be easier to understand what failed. Also I think it should be possible to build the table without the cloudquery binary.

yevgenypats (changes requested):
Looks good, just small comment on the test policy. Also once that working remember to move it to .github

.github/workflows/test_policy_sql.yml
disq: This file is mostly a copy of `test_policy.yml` (synced from [here](https://github.com/cloudquery/.github/blob/main/workflows/providers/test_policy.yml) which runs the current `cq policy` command testing) but instead we do the `psql` thing after creating the tables with `cq sync`.

.github/workflows/test_policy_sql.yml
disq: @yevgenypats done in [`f0ab33c` (#1220)](https://github.com/cloudquery/cq-provider-aws/pull/1220/commits/f0ab33c64a2d37d84fea5532133748a109a48ff5) (and [`7327b09` (#1220)](https://github.com/cloudquery/cq-provider-aws/pull/1220/commits/7327b094727ab0bebbb688db21a7ce29de825bf9))

.github/workflows/test_policy_sql.yml
```diff
@@ -0,0 +1,70 @@
...
+          restore-keys: |
+            ${{ runner.os }}-go-${{ matrix.go }}-
+
+      - name: Prepare for test - Create tables
+        run: |
+          go run ./test/gen-tables.go | psql -h localhost -p 5432 -U postgres -d postgres -w
+
+      - name: Run all policies
+        run: |
+          cd policies && psql -h localhost -p 5432 -U postgres -d postgres -w -f ./sql/policy.sql
```
yevgenypats, line 70: can we remove the `sql` folder?

yevgenypats (approved):
Looks good. just had one nit. Let's remove the `sql` folder (put everything directly under policies)?

PR #1255, opened 2022-07-20
feat: Add compliance dashboard

yevgenypats:
#### Summary

This also includes the following fixes:
* check field is a string
* execution_time is set only once and not in every subpolicy
* execution_time uses native PostgreSQL now() instead of data
which makes it work on windows
* make policies use relational paths and now you can execute it from any directory.

NOTE: There is still work to be done in the dashboard and potentially some rough edges but I think this is a good start.

Other ideas: add severity to checks which can make some of the visualizations more interesting.


shimonp21 (changes requested):
Why some places are 'execution_time' and some are 'execution_time'::timestamp?

disq (approved):

policies/README.md
```diff
@@ -30,3 +30,15 @@ psql -U postgres -c "select * from aws_policy_results" --csv
...
+
+Currently we have a pre-built compliance dashboard on top of `aws_policy_results` table which is available [here](../dashboards/grafana/compliance.json)
+
+<img alt="Azure Asset Inventory Grafana Dashboard" src="../dashboards/grafana/compliance.png" width=50% height=50%>
+
+### Instaltion
+
+1. Execute one or more of the above policies.
+2. Add the CloudQuery postgres database as a data source to Grafana (`Configuration -> Data Sources -> Add Data Source`)
+3. Import [dashboards/azure/azure_asset_inventory.json](../dashboards/grafana/compliance.json) into Grafana (`Import -> Upload JSON File`).
```
hermanschaaf, line 44: @yevgenypats I think we should move most of this `Visualization` section to a separate readme under the `dashboards` directory, and link to that section from here (maybe while keeping the screenshot). That way, you can look at the `dashboards` directory on its own and still know how to use it. WDYT?

hermanschaaf (approved):
Looks great :+1: I pushed some updates to the JSON to add a Data Source row and fix the colors of the distributions, which somehow always showed green in my case. Left a few questions, but nothing blocking

policies/README.md
yevgenypats: Yeah possible. I would still link to Visualization then from this README and also backlink from visualizations to policies.

yevgenypats:
> Why some places are 'execution_time' and some are 'execution_time'::timestamp?

some Postgres edge case in some queries where it couldn't do a cast implicitly 🤷 so it failed and I had to add that.

policies/README.md
hermanschaaf: Yup, that's what I had in mind with my comment too 👍

shimonp21 (approved):

PR #1383, opened 2022-08-04
feat: Update EC2 Services

bbernays:
#### Summary

Had to look into the EC2 instance resource from an unrelated reason. Saw that we could remove A LOT of custom code. spent a few minutes making the changes and opened a PR

- Add support for pagination in Security Groups
- Fix regional configs for BYO CIDR
- Update Resolvers to use new path resolver

---


resources/services/ec2/byoip_cidrs.go
```diff
@@ -68,14 +68,8 @@ func fetchEc2ByoipCidrs(ctx context.Context, meta schema.ClientMeta, parent *sch
...
 	// DescribeByoipCidrs does not work in next regions, so we ignore them.
 	if _, ok := map[string]struct{}{
-		"ap-northeast-2": {},
-		"af-south-1":     {},
 		"cn-north-1":     {},
 		"cn-northwest-1": {},
-		"eu-west-3":      {},
-		"eu-south-1":     {},
-		"eu-north-1":     {},
-		"me-south-1":     {},
```
bbernays, line 78: Support has been added for these regions!

resources/services/ec2/security_groups.go
```diff
@@ -230,24 +230,22 @@ func fetchEc2SecurityGroups(ctx context.Context, meta schema.ClientMeta, parent
...
+		})
+		if err != nil {
+			return diag.WrapError(err)
+		}
+		res <- output.SecurityGroups
+		if aws.ToString(output.NextToken) == "" {
+			break
+		}
+		config.NextToken = output.NextToken
 	}
```
bbernays, line 245: Support pagination

resources/services/ec2/byoip_cidrs.go
```diff
@@ -68,14 +68,8 @@ func fetchEc2ByoipCidrs(ctx context.Context, meta schema.ClientMeta, parent *sch
 	c := meta.(*client.Client)
 	// DescribeByoipCidrs does not work in next regions, so we ignore them.
 	if _, ok := map[string]struct{}{
-		"ap-northeast-2": {},
-		"af-south-1":     {},
 		"cn-north-1":     {},
 		"cn-northwest-1": {},
```
hermanschaaf, line 72: I wonder if there is a lesson we can draw from this for the future: maybe we should still make the call, but ignore the specific error that gets returned from unsupported regions? I don't know if that would have been possible though, necessarily.

hermanschaaf (approved):

resources/services/ec2/byoip_cidrs.go
bbernays: This was definitely an early way that we handled regional differences. I agree, today we usually would make the call and catch the specific error that would indicate the feature is not supported in the region. That being said the China and us-gov partitions we are not able to test in as we cannot get accounts in those partitions

PR #1385, opened 2022-08-04
feat: Simplify Resources

bbernays:
#### Summary

More simplification of resource resolvers

---


client/resolvers.go
```diff
@@ -29,6 +30,15 @@ func ResolveWAFScope(_ context.Context, meta schema.ClientMeta, r *schema.Resour
 	return diag.WrapError(r.Set(c.Name, meta.(*Client).WAFScope))
 }
 
+func ResolvePathJson(path string) func(_ context.Context, _ schema.ClientMeta, r *schema.Resource, c schema.Column) error {
+	return func(_ context.Context, _ schema.ClientMeta, r *schema.Resource, c schema.Column) error {
```
bbernays, line 34: Needs unit tests prior to being merged

hermanschaaf:
@bbernays it looks like the PR contains a couple of commits from another branch and will need to be rebased

hermanschaaf:
Maybe I'm missing something, but I think this is already possible. When using cq-gen, you can set the column type to be JSON. AFAICT It then uses `PathResolver` and does the JSON conversion for you. 

Here is a test case for this in cq-gen:
 - input: https://github.com/cloudquery/cq-gen/blob/main/codegen/tests/columns.hcl#L64-L72
 - and the expectation: https://github.com/cloudquery/cq-gen/blob/main/codegen/tests/expected/columns/change_type_default_resolver.go#L22-L26

So I'm not sure where `PathJSONResolver` would be needed?

bbernays:
I fully agree that if `PathResolver` can do the marshaling then this resolver is 100% not necessary. 
@roneli, @amanenk and @irmatov Can speak to why we have so many resolvers like this that as @hermanschaaf points out could possibly be handled by the `PathResolver`

https://github.com/cloudquery/cq-provider-aws/blob/01b56da6d86a41804ea7ddcce29946840041f9fc/resources/services/waf/web_acls.go#L254-L263

bbernays:
As @hermanschaaf has pointed out, this resolver is not necessary as our internal implementation for `json` columns already calls `json.marshal` 
https://github.com/cloudquery/cq-provider-sdk/blob/76ac8db01622b520bbacfe0272ab38233eb1abdf/provider/schema/dialect.go#L242-L251

amanenk:
It is totally fine to regenerate all the resources and use `PathResolver`. Before some sdk implementations not all the structures where automatically marshaled to json.

erezrokah (approved):
🧹

PR #1387, opened 2022-08-05
chore: Update codepipelines SDK code and use path resolvers where possible

hermanschaaf:
#### Summary

Replace custom code in codepipelines with CloudQuery SDK-provided resolvers.


hermanschaaf:
Wait... I just checked, and cq-gen will replace all this on the next generation. This is where it would be great to have the build check that regeneration yields no differences.

We can probably merge this anyway, but cq-gen should be updated to recognize these cases automatically.

hermanschaaf:
It was possible to update `gen.hcl` in such a way that the SDK resolvers get used. Re-generating now results in no differences :+1: It did require a small tweak to cq-gen, however: https://github.com/cloudquery/cq-gen/pull/122

resources/services/codepipeline/pipelines.go
```diff
@@ -12,19 +11,14 @@ import (
...
 //go:generate cq-gen --resource pipelines --config gen.hcl --output .
 func Pipelines() *schema.Table {
 	return &schema.Table{
 		Name:         "aws_codepipeline_pipelines",
-		Description:  "Represents the output of a GetPipeline action.",
+		Description:  "Represents the output of a GetPipeline action",
 		Resolver:     fetchCodepipelinePipelines,
 		Multiplex:    client.ServiceAccountRegionMultiplexer("codepipeline"),
-		IgnoreError:  client.IgnoreCommonErrors,
+		IgnoreError:  client.IgnoreAccessDeniedServiceDisabled,
```
bbernays, line 21: I think this needs to be updated in the `gen.hcl` file

hermanschaaf: Is `IgnoreCommonErrors` the one we want here?

hermanschaaf: Updated in the gen.hcl and regenerated to use `IgnoreCommonErrors` like it was before.

bbernays (approved):

resources/services/codepipeline/gen.hcl
```diff
@@ -1,11 +1,16 @@
...
 
+description_modifier "remove_read_only" {
+  words = ["  This member is required."]
+}
+
 resource "aws" "codepipeline" "pipelines" {
   path = "github.com/aws/aws-sdk-go-v2/service/codepipeline.GetPipelineOutput"
   ignoreError "IgnoreAccessDenied" {
-    path = "github.com/cloudquery/cq-provider-aws/client.IgnoreAccessDeniedServiceDisabled"
+    path = "github.com/cloudquery/cq-provider-aws/client.IgnoreCommonErrors"
```
candiduslynx, line 13: use `IgnoreCommonErrors` for the block name as well

candiduslynx (approved):

resources/services/codepipeline/gen.hcl
hermanschaaf: Done 👍

PR #1401, opened 2022-08-08
fix(build): Don't filter paths and enforce //check-for-changes on new cq-gen config files

hermanschaaf:
This updates the workflow that checks for generated code drift in two ways:
 - it is now always run, so that it can be made a required check. The `paths-filter` plugin ensures we stop early if no relevant files changed.
 - an additional check has been added to ensure that `//check-for-changes` is used in all new cq-gen config files


hermanschaaf:
Once approved, I will port this to the other provider repos as well (GCP, Azure, etc)

erezrokah (approved):

PR #1404, opened 2022-08-08
feat: Implement EC2 Key Pairs (#1403)

### LINKED ISSUE #1403: Support EC2 Key Pairs
### Describe the resource.

https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#EC2.DescribeKeyPairs

DescribeKeyPairs API operation for Amazon Elastic Compute Cloud.

Describes the specified key pairs or all of your key pairs.

### Use Case

Audit EC2 key pairs for deny-listed fingerprints. 

### Additional context

It might be good to have the `aws_ec2_instances` table also include the fingerprint. We can join on the key_name though. It might also be good to pull the public key material if there's a use case for it.

jhudson10x:
#### Summary

Add new resource for EC2 Key Pairs

https://github.com/cloudquery/cq-provider-aws/issues/1403

---

- [X] Read the [contribution guidelines](../blob/main/CONTRIBUTING.md) 🧑‍🎓
- [X] Run `go fmt` to format your code 🖊
- [X] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [X] Update or add tests. Learn more about testing [here](https://docs.cloudquery.io/docs/developers/sdk/testing) 🧪
- [X] Update the docs by running `go run ./docs/docs.go` and committing the changes 📃

### DIFF
--- a/resources/services/ec2/gen.hcl
+++ b/resources/services/ec2/gen.hcl
@@ -250,3 +250,43 @@ resource "aws" "ec2" "instance_types" {
     }
   }
 }
+
+resource "aws" "ec2" "key_pairs" {
+  path = "github.com/aws/aws-sdk-go-v2/service/ec2/types.KeyPairInfo"
+  ignoreError "IgnoreAccessDenied" {
+    path = "github.com/cloudquery/cq-provider-aws/client.IgnoreAccessDeniedServiceDisabled"
+  }
+  deleteFilter "AccountRegionFilter" {
+    path = "github.com/cloudquery/cq-provider-aws/client.DeleteAccountRegionFilter"
+  }
+  multiplex "AwsAccountRegion" {
+    path   = "github.com/cloudquery/cq-provider-aws/client.ServiceAccountRegionMultiplexer"
+    params = ["ec2"]
+  }
+
+  options {
+    primary_keys = ["key_pair_id"]
+  }
+
+  userDefinedColumn "account_id" {
+    description = "The AWS Account ID of the resource."
+    type        = "string"
+    resolver "resolveAWSAccount" {
+      path = "github.com/cloudquery/cq-provider-aws/client.ResolveAWSAccount"
+    }
+  }
+
+  userDefinedColumn "region" {
+    type        = "string"
+    description = "The AWS Region of the resource."
+    resolver "resolveAWSRegion" {
+      path = "github.com/cloudquery/cq-provider-aws/client.ResolveAWSRegion"
+    }
+  }
+
+  column "tags" {
+    type              = "json"
+    generate_resolver = false
+    description       = "The tags assigned to the key pair."
+  }
+}
--- a/resources/services/ec2/key_pairs.go
+++ b/resources/services/ec2/key_pairs.go
@@ -9,15 +9,16 @@ import (
 	"github.com/cloudquery/cq-provider-sdk/provider/schema"
 )
 
+//go:generate cq-gen --resource key_pairs --config gen.hcl --output .
 func Ec2KeyPairs() *schema.Table {
 	return &schema.Table{
 		Name:          "aws_ec2_key_pairs",
 		Description:   "Describes an EC2 Key Pair.",
 		Resolver:      fetchEc2KeyPairs,
 		Multiplex:     client.ServiceAccountRegionMultiplexer("ec2"),
-		IgnoreError:   client.IgnoreWithInvalidAction,
+		IgnoreError:   client.IgnoreAccessDeniedServiceDisabled,
 		DeleteFilter:  client.DeleteAccountRegionFilter,
-		Options:       schema.TableCreationOptions{PrimaryKeys: []string{"key_name"}},
+		Options:       schema.TableCreationOptions{PrimaryKeys: []string{"key_pair_id"}},
 		IgnoreInTests: true,
 		Columns: []schema.Column{
 			{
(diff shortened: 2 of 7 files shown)


resources/services/ec2/key_pairs.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/cloudquery/cq-provider-aws/client"
+	"github.com/cloudquery/cq-provider-sdk/provider/diag"
+	"github.com/cloudquery/cq-provider-sdk/provider/schema"
+)
+
+func Ec2KeyPairs() *schema.Table {
```
candiduslynx, line 12: Kindly add `cq-gen` config for this resource.
Additionally, the generate instruction comment should be present here, and no diff should occur once you run this generate command on this file.

See [AWS SES Template resource](https://github.com/cloudquery/cq-provider-aws/blob/main/resources/services/ses/templates.go#L13) for inspiraition.

resources/services/ec2/key_pairs.go
```diff
@@ -0,0 +1,78 @@
...
+	return &schema.Table{
+		Name:          "aws_ec2_key_pairs",
+		Description:   "Describes an EC2 Key Pair.",
+		Resolver:      fetchEc2KeyPairs,
+		Multiplex:     client.ServiceAccountRegionMultiplexer("ec2"),
+		IgnoreError:   client.IgnoreWithInvalidAction,
```
candiduslynx, line 18: Why did you choose this specific ignore function?
Will `client.client.IgnoreAccessDeniedServiceDisabled` work?

candiduslynx (changes requested):

resources/services/ec2/key_pairs.go
```diff
@@ -0,0 +1,78 @@
...
+		DeleteFilter:  client.DeleteAccountRegionFilter,
+		Options:       schema.TableCreationOptions{PrimaryKeys: []string{"key_name"}},
```
candiduslynx, line 20: Use `key_pair_id` instead

candiduslynx (changes requested):

resources/services/ec2/key_pairs.go
jhudson10x: I don't have a good reason for having chosen that specific ignore function. I will change it to client.IgnoreAccessDeniedServiceDisabled. If there is an even better ignore handler to use for this situation tell me and I will change it again.

resources/services/ec2/key_pairs.go
jhudson10x: I will add an entry to the `resources/services/ec2/gen.hcl` file for this resource. I notice that there are other resources with missing entries in this file. Can you provide me more info about the purpose of this file?

candiduslynx: We tend to use `cq-gen` to generate resources (may be installed via `make install-tools`).
Additionally, it provides an easy way to specify the resource structure.

Also, we are going to check for drift changes for the generated resources (so you should be able to run `go generate` on this file and see no changes whatsoever).

resources/services/ec2/gen.hcl
```diff
@@ -249,4 +249,44 @@ resource "aws" "ec2" "instance_types" {
...
+  multiplex "AwsAccountRegion" {
+    path   = "github.com/cloudquery/cq-provider-aws/client.ServiceAccountRegionMultiplexer"
+    params = ["ec2"]
+  }
+
+  options {
+    primary_keys = ["key_pair_id"]
+  }
+
+  userDefinedColumn "account_id" {
```
candiduslynx, line 271: does Key Pair have ARN?

candiduslynx: Then you should add a resolver for it, too.

Basically, you should add (and specify this column to e primary key, see lots of examples in the reources):
```hcl
userDefinedColumn "arn" {
    type              = "string"
    description       = "The Amazon Resource Name (ARN) of the Key Pair"
    generate_resolver = true
}
```
Then you'll need to regenerate the implementation and implement the newly generated resolver.
You can find examples of resolvers utilizing helper functions like [this one](https://github.com/cloudquery/cq-provider-aws/blob/main/client/helpers.go#L290) throughout the codebase.

jhudson10x: The func signature that cq-gen has a return type of `error`. Should I change the sig to return `schema.ColumnResolver` like the example you linked, or should I keep the generated sig and do like `resources/services/iot/iot_topic_rules.go#L1337` and write into the resource?

candiduslynx: You should do something like https://github.com/cloudquery/cq-provider-aws/blob/main/resources/services/ec2/images.go#L42-L44

I suppose you can just fill in the proper func body for the generated func.

erezrokah:
Hi @jhudson10x and @candiduslynx, I moved this PR to our monorepo (including original commits) via https://github.com/cloudquery/cloudquery/pull/1325. Since https://github.com/cloudquery/cloudquery/pull/1325 was approved and merged I suggest to do any follow up changes in another PR.
Thanks for the contribution @jhudson10x!

erezrokah:
Closing in favor https://github.com/cloudquery/cloudquery/pull/1325

PR #1396, opened 2022-08-07
feat: Add support for resource groups

adam-tylr:
#### Summary

Adds support for resource groups. 

---

Use the following steps to ensure your PR is ready to be reviewed

- [x] Read the [contribution guidelines](../blob/main/CONTRIBUTING.md) 🧑‍🎓
- [x] Run `go fmt` to format your code 🖊
- [x] Lint your changes via `golangci-lint run` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [x] Update or add tests. Learn more about testing [here](https://docs.cloudquery.io/docs/developers/sdk/testing) 🧪
- [x] Update the docs by running `go run ./docs/docs.go` and committing the changes 📃
- [ ] Ensure the status checks below are successful ✅

PR #1409, opened 2022-08-10
chore(tests): Disable decorder on resource files

hermanschaaf:
Decorder is a bit too strict on our cq-gen generated files, and is forcing us to place all structs at the top of the file, even though this is not the current behavior of cq-gen. It seems unnecessarily strict for generated files, let's disable it for now.

Relates to https://github.com/cloudquery/.github/issues/266

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -115,6 +115,7 @@ issues:
     - path: resources(\\|\/).*\.go
       linters:
         - dupl
+        - decorder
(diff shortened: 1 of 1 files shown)


hermanschaaf:
This should resolve the issue noted in https://github.com/cloudquery/cq-provider-aws/pull/1396

erezrokah (approved):

